Issue #36, opened 2020-04-26
Detected I/O inside the event loop


Golab:
Hello I’m on Home Assistant 0.109.0b3 and unifi ver: 
5.13.10 and protect ver: 1.13.3-beta1 and your addOn Is V0.1.6

My problem is that I get this error all the time in my log

Log Details (WARNING)
Logger: homeassistant.util.async_
Source: util/async_.py:120 
First occurred: 2:03:31 PM (6156 occurrences) 
Last logged: 4:36:42 PM

Detected I/O inside the event loop. This is causing stability issues. Please report issue to the custom component author for unifiprotect doing I/O at custom_components/unifiprotect/unifi_protect_server.py, line 109: verify=self._verify_ssl,
Detected I/O inside the event loop. This is causing stability issues. Please report issue to the custom component author for unifiprotect doing I/O at custom_components/unifiprotect/unifi_protect_server.py, line 210: verify=self._verify_ssl,

![image](https://user-images.githubusercontent.com/1252745/80310956-7d6c6780-87dd-11ea-8325-c227efd93ec9.jpeg)

![image](https://user-images.githubusercontent.com/1252745/80310964-852c0c00-87dd-11ea-86e1-9764d75fab33.jpeg)

![image](https://user-images.githubusercontent.com/1252745/80310972-8b21ed00-87dd-11ea-96a7-89221dd825c9.jpeg)

![image](https://user-images.githubusercontent.com/1252745/80311010-b1e02380-87dd-11ea-9736-d0262cced761.jpeg)

briis:
I am not yet on the 109 Beta, so I have not seen this error. The lines that are referenced are doing IO as these are the request calls to get the cameras and the events from the Protect Server. I will look in to this once I have a 109 version installed. But it seems that there are some changes that can severely impact this Integration.

Golab:
I really hop you can fix this sone it seems this error comes up at lest 10 times every minute, because if I clean log and wait 5 seconds and do a refresh then the error shows up again :/


If you have missed this


Detection of blocking I/O in the event loop

This is a bit technical, but some things happening in Home Assistant, like fetching data from an API or website, can lock up Home Assistant for a small moment, if not done correctly. Until now, this has been hard to detect.

This release adds logic in the core of Home Assistant that is able to detect if an integration does one of those blocking calls and writes a messages to the logs.


https://rc.home-assistant.io/blog/2020/04/21/release-109/

briis:
[...]

PR #39, opened 2020-05-02
Make component async

boralyl:
This fixes the warning that started in 0.109.0 (see #36 ) and converts the entire component to be async.  

I verified the following with these changes:
* Warning in HA 0.109 goes away.
* My switch/camera/binary sensor/sensors showed up correctly.
* All 3 registered services work.
* The `camera.snapshot` service works.

I apologize for the large diff, I tried a few simple things to get rid of the warning but ran into a lot of issues.  I decided to just go head and make it all async.

There are a lot of formatting changes due to my IDE auto-formatting code using [black](https://black.readthedocs.io/en/stable/) standards (which HA also uses in their code base).

### DIFF
--- a/custom_components/unifiprotect/__init__.py
+++ b/custom_components/unifiprotect/__init__.py
@@ -129,7 +129,7 @@ async def async_setup(hass: core.HomeAssistant, config: dict) -> bool:
         upv_server = upv.UpvServer(
             session, host, port, username, password, use_ssl, minimum_score
         )
-        _LOGGER.error("Connected to Unifi Protect Platform")
+        _LOGGER.debug("Connected to Unifi Protect Platform")
     except upv.NotAuthorized:
         _LOGGER.error("Authorization failure while connecting to NVR")
         return False
@@ -139,7 +139,6 @@ async def async_setup(hass: core.HomeAssistant, config: dict) -> bool:
     except requests.exceptions.ConnectionError as ex:
         _LOGGER.error("Unable to connect to NVR: %s", str(ex))
         raise PlatformNotReady
-    _LOGGER.error("upv_server: %s", upv_server)
 
     coordinator = DataUpdateCoordinator(
         hass,
@@ -150,41 +149,40 @@ async def async_setup(hass: core.HomeAssistant, config: dict) -> bool:
     )
     # Fetch initial data so we have data when entities subscribe
     await coordinator.async_refresh()
-    _LOGGER.error("async refreshed.")
     hass.data[UPV_DATA] = {
         "coordinator": coordinator,
         "upv": upv_server,
     }
 
-    # async def async_save_thumbnail(call):
-    #    """Call save video service handler."""
-    #    await async_handle_save_thumbnail_service(hass, call)
+    async def async_save_thumbnail(call):
+        """Call save video service handler."""
+        await async_handle_save_thumbnail_service(hass, call)
 
-    # async def async_set_recording_mode(call):
-    #    """Call Set Recording Mode."""
-    #    await async_handle_set_recording_mode(hass, call)
+    async def async_set_recording_mode(call):
+        """Call Set Recording Mode."""
+        await async_handle_set_recording_mode(hass, call)
 
-    # async def async_set_ir_mode(call):
-    #    """Call Set Infrared Mode."""
-    #    await async_handle_set_ir_mode(hass, call)
+    async def async_set_ir_mode(call):
+        """Call Set Infrared Mode."""
+        await async_handle_set_ir_mode(hass, call)
 
-    # hass.services.register(
-    #    DOMAIN,
-    #    SERVICE_SAVE_THUMBNAIL,
-    #    async_save_thumbnail,
-    #    schema=SAVE_THUMBNAIL_SCHEMA,
-    # )
+    hass.services.async_register(
+        DOMAIN,
+        SERVICE_SAVE_THUMBNAIL,
+        async_save_thumbnail,
+        schema=SAVE_THUMBNAIL_SCHEMA,
+    )
 
-    # hass.services.register(
-    #    DOMAIN,
-    #    SERVICE_SET_RECORDING_MODE,
-    #    async_set_recording_mode,
-    #    schema=SET_RECORDING_MODE_SCHEMA,
-    # )
+    hass.services.async_register(
+        DOMAIN,
+        SERVICE_SET_RECORDING_MODE,
+        async_set_recording_mode,
+        schema=SET_RECORDING_MODE_SCHEMA,
+    )
 
-    # hass.services.register(
-    #    DOMAIN, SERVICE_SET_IR_MODE, async_set_ir_mode, schema=SET_IR_MODE_SCHEMA,
-    # )
+    hass.services.async_register(
+        DOMAIN, SERVICE_SET_IR_MODE, async_set_ir_mode, schema=SET_IR_MODE_SCHEMA,
+    )
 
     return True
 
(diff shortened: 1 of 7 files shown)


custom_components/unifiprotect/__init__.py
```diff
@@ -119,13 +123,13 @@ def setup(hass, config):
     use_ssl = conf.get(CONF_SSL)
     minimum_score = conf.get(CONF_MIN_SCORE)
     scan_interval = conf[CONF_SCAN_INTERVAL]
+    session = async_get_clientsession(hass)
```
boralyl, line 126: This `asyncio.ClientSession` will be used to make async requests in the `UpvServer` class.

custom_components/unifiprotect/__init__.py
```diff
@@ -136,6 +140,20 @@ def setup(hass, config):
         _LOGGER.error("Unable to connect to NVR: %s", str(ex))
         raise PlatformNotReady
```
boralyl, line 142: This is the preferred way by HA to fetch data when it's used for many entities: https://developers.home-assistant.io/docs/integration_fetching_data#coordinated-single-api-poll-for-data-for-all-entities

custom_components/unifiprotect/camera.py
```diff
@@ -134,35 +135,29 @@ def update(self):
...
+    async def async_disable_motion_detection(self):
         """Disable motion detection in camera."""
-        ret = self.upv_object.set_camera_recording(self._camera_id, "never")
+        ret = await self.upv_object.set_camera_recording(self._camera_id, "never")
         if not ret:
             return
 
         self._motion_status = "never"
         self._isrecording = False
         _LOGGER.debug("Motion Detection Disabled for Camera: %s", self._name)
```
boralyl, line 157: This is no longer needed, the async method below will be used.

custom_components/unifiprotect/unifi_protect_server.py
```diff
@@ -1,10 +1,12 @@
 """Unifi Protect Server Wrapper."""
```
boralyl, line 1: The main change was altering this class to make async requests.

custom_components/unifiprotect/unifi_protect_server.py
```diff
@@ -40,293 +53,248 @@ def __init__(
         self.device_data = {}
 
         urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
-        self.req = requests.session()
-        self._api_auth_bearer_token = self._get_api_auth_bearer_token()
```
boralyl, line 44: I had to move these 2 calls out of __init__ as you can do anyc calls in the an `__init__` method.  The api_auth_bearer_token will get set the first time HA calls the `update` method on this class.

custom_components/unifiprotect/unifi_protect_server.py
```diff
@@ -40,293 +53,248 @@ def __init__(
...
-    def update(self):
+    async def update(self) -> dict:
         """Updates the status of devices."""
-        self._get_camera_list()
-        self._get_motion_events(10)
+        if self._api_auth_bearer_token is None:
+            self._api_auth_bearer_token = await self._get_api_auth_bearer_token()
+        await self._get_camera_list()
+        await self._get_motion_events(10)
+        return self.devices
```
boralyl, line 70: This is what will be returned in your sensors and switches when you see references to `coordinator.data`.

custom_components/unifiprotect/unifi_protect_server.py
```diff
@@ -40,293 +53,248 @@ def __init__(
...
 
-    def _get_api_auth_bearer_token(self):
+    async def _get_api_auth_bearer_token(self) -> str:
         """get bearer token using username and password of local user."""
-
```
boralyl, line 59: I cleaned up a bunch of these strings to use `f` strings and use a common base_url instead of repeating it in all of the methods.

custom_components/unifiprotect/unifi_protect_server.py
```diff
@@ -40,293 +53,248 @@ def __init__(
...
-            + str(self._port)
-            + "/api/events?end="
-            + str(end_time)
-            + "&start="
-            + str(start_time)
-            + "&type=motion"
-        )
-
-        response = self.req.get(
+        event_uri = f"{self._base_url}/api/events"
```
boralyl, line 194: Also updated some of these GET requests to pass the params, rather than putting them in the url query string.  It's a bit safer this way as they will get encoded properly by aiohttp.

briis:
Hi Guys,
Fantastic work you did, and I can confirm that everything also loads up here, without any errors. 
During the testing I still found a few issues.
* Binary Sensors for Motion detection do not trigger, even with recording set to motion or always
* The switch component shows up correctly, but I cannot flip any of the switches. They just go back to previous state.

Are you seeing the same things?

boralyl:
@briis  Thanks for testing this out. I'll take a look at those issues tomorrow, it's very likely I missed something as I only use part of the functionality of this component in my own setup.

custom_components/unifiprotect/switch.py
```diff
@@ -71,50 +78,59 @@ async def async_setup_platform(hass, config, async_add_entities, _discovery_info
...
         self._switch_type = SWITCH_TYPES.get(switch_type)[2]
         _LOGGER.debug("UnifiProtectSwitch: %s created", self._name)
-        _LOGGER.debug("UnifiProtectSwitch: IR_ON %s IR_OFF %s", self._ir_on_cmd, self._ir_off_cmd)
+        _LOGGER.debug(
+            "UnifiProtectSwitch: IR_ON %s IR_OFF %s", self._ir_on_cmd, self._ir_off_cmd
+        )
 
     @property
     def should_poll(self):
         """Poll for status regularly."""
```
boralyl, line 115: When using a data coordinator this should be false, as this entity isn't responsible for updating the data.

custom_components/unifiprotect/switch.py
```diff
@@ -132,36 +148,31 @@ def device_state_attributes(self):
...
         """Turn the device off."""
         if self._switch_type == "ir_mode":
             _LOGGER.debug("Turning off IR")
-            self.data.set_camera_ir(self._camera_id, self._ir_off_cmd)
+            await self.upv.set_camera_ir(self._camera_id, self._ir_off_cmd)
         else:
             _LOGGER.debug("Turning off Recording")
-            self.data.set_camera_recording(self._camera_id, TYPE_RECORD_NEVER)
-
-    def update(self):
```
boralyl, line 156: The data coordinator handles updating the data so this shouldn't be implemented.  I moved the logic in here to the is_on property.

boralyl:
> Hi Guys,
> Fantastic work you did, and I can confirm that everything also loads up here, without any errors.
> During the testing I still found a few issues.
> 
>     * Binary Sensors for Motion detection do not trigger, even with recording set to motion or always
> 
>     * The switch component shows up correctly, but I cannot flip any of the switches. They just go back to previous state.
> 
>     * If I change the recording status for a camera within he Unifi Protect App, the new setting does not get reflected in Home Assistant. However if done by calling the service `unifiprotect.set_recording_mode` the new setting is reflected.
> 
> 
> Without having scrutinised all the code it seems that the background update routine is not working?
> 
> Are you seeing the same things?

@briis Ok, just tested all of the sensors/switches/binary_sensors out locally.  Initially I was seeing the same problem you were, but I believe I've got them all fixed now.

briis:
YOU ROCK - Everything works now, and no errors in the log. I can't thank you enough - I have learned a lot by looking at the code, and now it works. THANK YOU.

While you were doing all this, I looked at making it in to a real integration, so that it can be configured from the Interface, so that will be next. And then supporting the UnifiOS.

I will be merging this PR now, and release it to the public.

THANK YOU.
/B

B-Hartley:
Thanks guys !
Really pleased to see this modified so quickly !

I might be imagining it, but it seems to be working visibly quicker.

Thank you !!!

PR #45, opened 2020-05-07
DevContainer: Enable debugger 

msvinth:
Hi Bjarne,

Just a small PR to improve the .devcontainer setup so that the debugger in VS Code works to debug the integration.

Also note that when you use create the devcontainer it pulls the latests version of the Home Assistant dev branch. And I get some error right now during init as the dev branch seems to some error currently.
In the VS code shell you can type "container set-version" to chose which Home Assistant version is used e.g. 0.109.4
Also if you type just "container" it will list the commands available.

### DIFF
--- a/.devcontainer/configuration.yaml
+++ b/.devcontainer/configuration.yaml
@@ -1,6 +1,9 @@
 default_config:
 cloud:
 
+ptvsd:
+  wait: False # Set to True if you want hass to break and wait for debugger to attach
+
 stream:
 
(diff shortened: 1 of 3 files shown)


briis:
Thanks. I have spend some time today playing with the devcontainer, and even though there are still some things I am not completely sure about, it is really a strong tool, and much faster to develop and to test in. Thanks a lot for the hints. Changes are merged.

PR #71, opened 2020-05-27
Fix event_score attribute

thegame3202:


briis:
There you go, many changes, and then this messed up. Thanks for spotting.

Issue #97, opened 2020-08-06
Use WebSockets for events


adrum:
Unifi Protect apparently exposes a WebSocket interface that would allow for real-time updates of events such as motion and doorbell button events. 

This was recently implemented by this library: https://github.com/hjdhjd/homebridge-unifi-protect2

It would be amazing if we could implement real-time events.

briis:
Thanks for spotting this. Apparently the WebSocket support is only for UnifiOS devices, so UDMP and UNVR and not the CloudKey, but never the less, this looks very interesting. I will put it on the to-look-at list, and see if I can figure out a way to implement this. Anyone else who wants to take a shot at this, feel free to jump in.

adrum:
@briis I've started work on this. I have the connection setup, just parsing the response now.

briis:
Fantastic, I will wait for what you come up with. 
Thanks

mcondren:
Is there any way this could be used to ‘insert’ a motion event flag? I have the rtsp streams from protect running into Frigate [https://github.com/blakeblackshear/frigate](url) for person detection. Would be awesome if when Home Assistant detected motion on these streams (either via frigate integration or built-in tensorflow integration) it could trigger a motion event in protect. This would make scrolling through the timeline to find a detection much easier as it highlights motion events.

OmenWild:
> Is there any way this could be used to ‘insert’ a motion event flag?

Like +Infinity on this. I would love to let my desktop do person detection to trigger the motion event on the Unifi side. So. Many. Spider. Webs.

briis:
Hi @mcondren and @OmenWild 
This sounds like an interesting use-case, but before I look at that, we need to get the WebSocket implementation done. @adrum, has done the initial work on the IO component, but at present it has not been implemented in the Integration itself. I am honestly not completely sure how to do that, so it might take a little while. But once it is done, I will look at your request.

mcondren:
Thanks @briis

PR #124, opened 2020-10-24
Add websocket support

bdraco:
Motion and doorbell updates should now happen
right away

Reduces the amount of entity updates since we
now only update cameras that change when we
poll instead of them all.

Fixes https://github.com/briis/unifiprotect/issues/97

Requires https://github.com/briis/pyunifiprotect/pull/6
Thanks to @hjdhjd for decoding the protocol

### DIFF
--- a/custom_components/unifiprotect/__init__.py
+++ b/custom_components/unifiprotect/__init__.py
@@ -1,13 +1,12 @@
 """Unifi Protect Platform."""
 
 import asyncio
-from datetime import timedelta
 import logging
+from datetime import timedelta
 
+import homeassistant.helpers.device_registry as dr
 from aiohttp import CookieJar
 from aiohttp.client_exceptions import ServerDisconnectedError
-from pyunifiprotect import NotAuthorized, NvrError, UpvServer
-
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import (
     CONF_HOST,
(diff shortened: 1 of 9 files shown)


bdraco:
Looks like I've introduced a bug in unloading.  Marking it as a draft to fix that

bdraco:
All fixed. Verified I could add and remove the integration without error and it disconnects the websocket

briis:
THANK YOU. I will update the manifest to work with the updated pyunifiprotect 0.21.0

briis:
When running the new version, motion detection works perfectly, but I get the following error when trying to change the recording mode from the Switch. Basically all the Switches fail with the same error:

```python
'UnifiProtectData' object has no attribute 'async_request_refresh'
  File "/usr/lib/python3.8/site-packages/homeassistant/components/websocket_api/commands.py", line 138, in handle_call_service
    await hass.services.async_call(
  File "/usr/lib/python3.8/site-packages/homeassistant/core.py", line 1408, in async_call
    task.result()
  File "/usr/lib/python3.8/site-packages/homeassistant/core.py", line 1443, in _execute_service
    await handler.job.target(service_call)
  File "/usr/lib/python3.8/site-packages/homeassistant/helpers/entity_component.py", line 204, in handle_service
    await self.hass.helpers.service.entity_service_call(
  File "/usr/lib/python3.8/site-packages/homeassistant/helpers/service.py", line 470, in entity_service_call
    future.result()  # pop exception if have
  File "/usr/lib/python3.8/site-packages/homeassistant/helpers/entity.py", line 664, in async_request_call
    await coro
  File "/usr/lib/python3.8/site-packages/homeassistant/helpers/service.py", line 507, in _handle_entity_call
    await result
  File "/config/custom_components/unifiprotect/switch.py", line 142, in async_turn_on
    await self.protect_data.async_request_refresh()
AttributeError: 'UnifiProtectData' object has no attribute 'async_request_refresh'
````

bdraco:
I missed updating the method.  I'll send a PR in a moment to fix it

briis:
The fix removed the error, but the switches don't work. They flip for a second and then return to previous state - and the state does not change. I tested the Service itself, from the developer menu, and that works. And when doing it from here, the Switch also changes to the correct position. Any ideas?

bdraco:
For me the switches are updating but since there camera update isn't ready yet they don't change state for 60 seconds.

I'm working on improving that now

briis:
You are right - there is a delay, but it works. I was a bit impatient

PR #125, opened 2020-10-24
Fix missed method refactor in switch

bdraco:


bdraco:
I noticed that since we delay the camera updates to every 60 seconds, it doesn't update right away.  I'll fix that as well

PR #126, opened 2020-10-24
Force a camera update when changing the switch

bdraco:
Force a camera update when changing the switch so it updates right away.

Disconnect websocket when the integration is unloaded.

Requires https://github.com/briis/pyunifiprotect/pull/7


bdraco:
Good to go now, needs the version bump from https://github.com/briis/pyunifiprotect/pull/7

briis:
Smart way of fixing this.

briis:
Works like a charm. I will write up some documentation tomorrow (Its night here) and then make a Beta release available in HACS.
Once again, a BIG thank you for helping out.

PR #130, opened 2020-10-26
Ensure the config entry listner is unsubscribed on unload

bdraco:
I noticed the listener wasn't being removed when the integration was removed.

Small tweaks

### DIFF
--- a/custom_components/unifiprotect/__init__.py
+++ b/custom_components/unifiprotect/__init__.py
@@ -1,12 +1,13 @@
 """Unifi Protect Platform."""
 
 import asyncio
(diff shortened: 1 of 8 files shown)


briis:
Thanks, and you cleaned up the code also :)

briis:
Last question: I guess we are now close to being able to add this as core HA Integration. I read through [this guide](https://developers.home-assistant.io/docs/creating_component_code_review) and I have a Fork of HA running, but I am not completely sure how to add the necessary tests. Do you have a pointer to where I can read up on this?

PR #172, opened 2021-01-07
Update services.yaml

MarcJenningsUK:
Added a missing quote to allow for example data to show in the UI


briis:
Thank you. I had missed that.

PR #183, opened 2021-01-21
Add a retry if the first update request fails

### LINKED ISSUE #181: return self._device_data["event_on"]   KeyError: 'event_on'
2021-01-17 16:02:31 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved

  File "/config/custom_components/unifiprotect/data.py", line 47, in async_refresh

    self._async_process_updates(

  File "/config/custom_components/unifiprotect/data.py", line 63, in _async_process_updates

    self.async_signal_device_id_update(device_id)

  File "/config/custom_components/unifiprotect/data.py", line 96, in async_signal_device_id_update

    update_callback()

  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 296, in async_write_ha_state

    self._async_write_ha_state()

  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 320, in _async_write_ha_state

    sstate = self.state

  File "/usr/src/homeassistant/homeassistant/components/binary_sensor/__init__.py", line 158, in state

    return STATE_ON if self.is_on else STATE_OFF

  File "/config/custom_components/unifiprotect/binary_sensor.py", line 92, in is_on

    return self._device_data["event_on"]

KeyError: 'event_on'

bdraco:
We would continue on if the first bootstrap call failed which could lead to partial setup. Usually the 500 error is transient so throwing `ConfigEntryNotReady` will cause the config entry to try again later.

Fixes #181

### DIFF
--- a/custom_components/unifiprotect/__init__.py
+++ b/custom_components/unifiprotect/__init__.py
@@ -86,6 +86,8 @@ async def async_setup_entry(hass: HomeAssistantType, entry: ConfigEntry) -> bool
         raise ConfigEntryNotReady from notreadyerror
 
     await protect_data.async_setup()
+    if not protect_data.last_update_success:
+        raise ConfigEntryNotReady
 
     update_listener = entry.add_update_listener(_async_options_updated)
 


bdraco:
Actually I found another issue, but I'm out of time for today. I'll circle back when I have a fix.

briis:
Hi Nick,
I found and fixed an error with the Set Status Light service, that I want to bring out - should I add this fix now, or are you still planning to fix the other issue you found?

bdraco:
> Hi Nick,
> I found and fixed an error with the Set Status Light service, that I want to bring out - should I add this fix now, or are you still planning to fix the other issue you found?

I've been swamped so I haven't had a chance to investigate this further.

It probably makes sense to merge this as-is and then do a followup when I have time since it will at least help some of the cases.

briis:
Thanks Nick, its merged.

PR #191, opened 2021-02-01
Use the server id for the unique id

bdraco:
The unique id needs to be stable.  Since the name can change, we should use the device_id since it never will

needs https://github.com/briis/pyunifiprotect/pull/22


briis:
Thanks. I have updated pyunifiprotect and will change the version number in the manifest.

PR #213, opened 2021-03-21
fix(g4doorbell/chime_duration): Add missing SCHEMA

mohlek:


mohlek:
Sry missed that.
Still not tested. Need to prepare test environment first

briis:
No worries, I had already made this change, and I will test it in a minute.
There was also something missing in camera.py, which I have fixed.

mohlek:
looking forward to see this being released.
Thanks for your work.

briis:
I am more or less done now. I decided to make the step interval 100 instead of 1, on the duration. 
But just download the changed files, overwrite what you have in the custom_components directory and you should be up and running now.

Thanks for your help.

mohlek:
works as intended. Thanks

Issue #217, opened 2021-04-04
ClientSession is being leaked on setup retry/auth failure


bdraco:
If the integration fails to setup we need to manually close the client session created from 
`async_create_clientsession` before returning/raising

bdraco:
I'll put something together when I've had more sleep

briis:
Thanks Nick.

bdraco:
This affects a lot of integrations 
https://github.com/home-assistant/core/pull/47886#issuecomment-800611244

I'm going to do an audit. It will be a few days

briis:
It looks like a need to make some changes to other Integrations I made as well, once I see how you fix this. 😅

bdraco:
Potential solution https://github.com/home-assistant/core/pull/48908

briis:
I looked at the code, and will this then fix it for all Integrations, without any code change?

Btw: Isn't it very late where you are  😌

bdraco:
> I looked at the code, and will this then fix it for all Integrations, without any code change?

Thats the idea
> 
> Btw: Isn't it very late where you are  😌

...and yes it is

bdraco:
The change is happening in next month's release so the problem will just go away :)

Anyone who has the leak probably just needs to restart once or twice a month to avoid the memory exhaustion until next month's release.

briis:
Perfect. Thanks for the work you put in for all of us to fix this.

PR #220, opened 2021-04-09
Return False when auth fails

bdraco:
Fixes error when auth failes `2021-04-09 03:16:36 ERROR (MainThread) [homeassistant.config_entries] unifiprotect.async_setup_entry did not return boolean`


briis:
This is the fix for the issue #217 you opened, correct?

bdraco:
Nope, just something I noticed in the process of testing #217

briis:
Ok, thanks.

PR #221, opened 2021-04-13
Stop updates on stop event to prevent shutdown delay

bdraco:


bdraco:
Tested by reloading and restarting.

Should probably get some more testing before merge.

PR #224, opened 2021-04-23
Fix compat with latest dev

bdraco:
HomeAssistantType was removed in https://github.com/home-assistant/core/pull/49593

### DIFF
--- a/custom_components/unifiprotect/__init__.py
+++ b/custom_components/unifiprotect/__init__.py
@@ -8,18 +8,19 @@ from aiohttp import CookieJar
 from aiohttp.client_exceptions import ServerDisconnectedError
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import (
-    EVENT_HOMEASSISTANT_STOP,
     CONF_HOST,
     CONF_ID,
     CONF_PASSWORD,
     CONF_PORT,
     CONF_SCAN_INTERVAL,
     CONF_USERNAME,
+    EVENT_HOMEASSISTANT_STOP,
 )
+from homeassistant.core import HomeAssistant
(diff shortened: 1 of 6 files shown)


briis:
Thanks Nick.
Just tested this on a 2021.4 system, and here it does not work. Is that expected?
I guess I will need to change the HACS file to ensure that only people with 2021.5 + version install this.

bdraco:
I tried it with 2021.4.6 and it worked for me. Which error did you see?

briis:
On purpose I did not update my DEV container, which was still running 2021.4.dev. 
I do not get errors, but it loops when trying to connect to the UDMP. I then removed the Integration setup and tried to add it again, it gives me a `Cannot connect to host 192.168.1.1:443` error in the Config Flow.

I have a system still running 2021.4.1 where I tried the previous version, and I can connect from there with the same credentials. I will now try and add the updated files from this pull to that system. If that works, then it is most likely just my DEV container that sucks.

briis:
Ok, just updated with this pull on my other test system running 2021.4.1 and it works perfect. So it must be my dev container that needs updating.

briis:
It still works, except for my Devcontainer (I use ghcr.io/ludeeus/devcontainer/integration:stable). I deleted the integration and when trying to set it up again, I keep getting `Error requesting data from 192.168.1.1: Cannot connect to host 192.168.1.1:443`

It is so strange as this has worked flawlessly before. I guess only change is a new docker version on my Mac.

briis:
Ok. Removed the latest version of Docker from my Mac Dev machine, and installed a release that was a month old. Now all works. Thanks for the initial fix Nick. I just found I had a few others to do, with this change.

PR #226, opened 2021-05-04
Added helpful tip for debug logging

blaines:
Just thought this would be useful for folks


briis:
Thank you, and you are right, this should have been in there a long time ago.

PR #252, opened 2021-06-20
Clarify Configuration paragraph

marcusrugger:
I know this paragraph was probably perfectly clear to HA veterans, but for a noob like me who is still learning my way around, this little bit of clarification would have eliminated some confusion on my part.


briis:
Thanks.

PR #328, opened 2021-11-08
Fixing Standard Issue

briis:
I ran through your suggestions Nick, and have modified `config_flow.py` and `__init__.py` according to your suggestions. It does work, but I am getting an error when I run the STOP command from the Server Management page, and I am not sure why that happens. Here is the log:

```log
2021-11-08 07:06:53 ERROR (MainThread) [homeassistant] Error doing job: Exception in callback UnifiProtectData.async_stop(<Event homeassistant_stop[L]>)
  File "/usr/local/python/lib/python3.9/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/local/python/lib/python3.9/site-packages/homeassistant/core.py", line 835, in _onetime_listener
    self._hass.async_run_job(listener, event)
  File "/usr/local/python/lib/python3.9/site-packages/homeassistant/core.py", line 452, in async_run_job
    return self.async_run_hass_job(HassJob(target), *args)
  File "/usr/local/python/lib/python3.9/site-packages/homeassistant/core.py", line 436, in async_run_hass_job
    return self.async_add_hass_job(hassjob, *args)
  File "/usr/local/python/lib/python3.9/site-packages/homeassistant/core.py", line 363, in async_add_hass_job
    task = self.loop.create_task(hassjob.target(*args))
TypeError: async_stop() takes 1 positional argument but 2 were given
````

### DIFF
--- a/custom_components/unifiprotect/__init__.py
+++ b/custom_components/unifiprotect/__init__.py
@@ -41,10 +41,10 @@ _LOGGER = logging.getLogger(__name__)
 SCAN_INTERVAL = timedelta(seconds=DEFAULT_SCAN_INTERVAL)
 
 
-async def async_setup(hass: HomeAssistant, config: ConfigType) -> bool:
-    """Set up the Unifi Protect components."""
+# async def async_setup(hass: HomeAssistant, config: ConfigType) -> bool:
+#     """Set up the Unifi Protect components."""
 
-    return True
+#     return True
 
 
 async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: custom_components/unifiprotect/__init__.py [hunk 1]
```diff
@@ -93,31 +87,23 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
...
 
-    for platform in UNIFI_PROTECT_PLATFORMS:
-        hass.async_create_task(
-            hass.config_entries.async_forward_entry_setup(entry, platform)
-        )
+    hass.config_entries.async_setup_platforms(entry, UNIFI_PROTECT_PLATFORMS)
+
+    entry.async_on_unload(entry.add_update_listener(_async_options_updated))
+    entry.async_on_unload(
+        hass.bus.async_listen_once(EVENT_HOMEASSISTANT_STOP, protect_data.async_stop)
```
bdraco, line 105: When the callback runs from `async_listen_once` it will call `async_stop` with the stop event as the first argument

protect_data.async_stop(event)

If you change the signature of `async_stop` to something like `async_stop(self, *args)` it will not throw the error anymore.

custom_components/unifiprotect/config_flow.py
```diff
@@ -92,17 +92,13 @@ async def async_step_user(self, user_input=None):
...
             },
             options={
                 CONF_USERNAME: user_input.get(CONF_USERNAME),
                 CONF_PASSWORD: user_input.get(CONF_PASSWORD),
-                CONF_DISABLE_RTSP: user_input.get(CONF_DISABLE_RTSP),
-                CONF_DOORBELL_TEXT: user_input.get(CONF_DOORBELL_TEXT),
-                CONF_SNAPSHOT_DIRECT: user_input.get(CONF_SNAPSHOT_DIRECT),
+                CONF_DISABLE_RTSP: False,
+                CONF_DOORBELL_TEXT: "",
+                CONF_SNAPSHOT_DIRECT: False,
```
bdraco, line 101: We need to migrate the data in `__init__.py` `async_setup_entry` using `async_update_entry` to move it from `data` into `options` and then change the places where the data is looked up in the integration

There is an example in the yeelight integration see `_async_normalize_config_entry`

custom_components/unifiprotect/__init__.py
briis: Thanks, this did the trick.

custom_components/unifiprotect/config_flow.py
briis: Thanks. I will have a look and make the change.

PR #331, opened 2021-11-09
Migrated to extra_state_attributes

briis:
All entities now using `extra_state_attributes` and all comments in these files have been removed.

### DIFF
--- a/custom_components/unifiprotect/camera.py
+++ b/custom_components/unifiprotect/camera.py
@@ -3,11 +3,7 @@ import logging
 from typing import Optional
 
 from homeassistant.components.camera import SUPPORT_STREAM, Camera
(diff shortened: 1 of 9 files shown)


custom_components/unifiprotect/camera.py
```diff
@@ -206,24 +192,24 @@ def is_recording(self):
         )
 
     @property
-    def device_state_attributes(self):
+    def available(self):
+        """Return if entity is available."""
+        return self._device_data["online"]
```
bdraco, line 197:
```suggestion
        return self._device_data["online"] and super().available
```

bdraco (commented):
Looks good.  One small suggestion above

PR #339, opened 2021-11-17
Switch to a dataclass for config entry data

bdraco:


briis:
Is this in preparation for switching to the `ProtectApiClient` from `UpvServer`?

bdraco:
Nope, its just a cleanup so its easier to type the data stored in `hass.data` later.

bdraco:
I still need to find some type to finish up the cleanups to `select.py` and implement a reauth flow. Once thats done I think we are pretty close to ready for a core merge

briis:
Thanks a lot Nick.
Shouldn't we move the the new ProtectApiClient also before the core integration? I don't understand all that has been done, but it looks smart.

When we get to that, I do need your help with the migration of the Websocket setup, as this is completely different from what you made in `data.py`.

bdraco:
> Shouldn't we move the the new ProtectApiClient also before the core integration? I don't understand all that has been done, but it looks smart.

Yes we should.  I think that is the last step though since it would be good to get all the cleanups done first.  I have a few more PRs after this one.

AngellusMortis:
I was able to successfully add the UFP integration and reconfigure it. It looks like that is where the majority of the changes were. Anything in particular you want me to test?

AngellusMortis (approved):

bdraco:
> I was able to successfully add the UFP integration and reconfigure it. It looks like that is where the majority of the changes were. Anything in particular you want me to test?

This one likely blows up or it doesn't. Loading enough is probably enough testing.

Thanks 👍

PR #340, opened 2021-11-17
Select cleanup

bdraco:
First round of select cleanups

### DIFF
--- a/custom_components/unifiprotect/select.py
+++ b/custom_components/unifiprotect/select.py
@@ -7,6 +7,7 @@ import logging
 from homeassistant.components.select import SelectEntity, SelectEntityDescription
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.core import HomeAssistant
+from homeassistant.exceptions import HomeAssistantError
 
 from .const import (
     CUSTOM_MESSAGE,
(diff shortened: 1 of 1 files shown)


custom_components/unifiprotect/select.py
```diff
@@ -49,11 +51,24 @@
...
 LIGHT_MODE_MOTION = "On Motion"
 LIGHT_MODE_DARK = "When Dark"
 LIGHT_MODE_OFF = "Manual"
 LIGHT_MODES = [LIGHT_MODE_MOTION, LIGHT_MODE_DARK, LIGHT_MODE_OFF]
 
+LIGHT_MODE_TO_SETTINGS = {
+    LIGHT_MODE_MOTION: (TYPE_LIGHT_RECORD_MOTION, "fulltime"),
+    LIGHT_MODE_DARK: (TYPE_RECORD_ALWAYS, "dark"),
+    LIGHT_MODE_OFF: (TYPE_RECORD_OFF, None),
+}
```
AngellusMortis, line 64: Do you think it would be a good time to start using the enums from pyunifiprotect? 

There should be enums in there for all of the selects.

https://github.com/briis/pyunifiprotect/blob/master/pyunifiprotect/data/types.py#L142

bdraco: It would be nice to do in the future as a code quality improvement.

Right now I've tried to limit changes to what would come up in a core so it can merge quickly.

AngellusMortis: Far enough. All of the selects look good from testing.

AngellusMortis (approved):

custom_components/unifiprotect/select.py
bdraco: Thanks!

PR #348, opened 2021-11-24
Fixes HLS streaming

AngellusMortis:
* Tweaks dev configuration a bit more
* Adds VS Code task to easily install dev version of HA
* Fixes HLS streaming for camera entities

While I was testing stuff for @bdraco, I noticed that I could no longer do streaming with my UFP cameras. After digging into it, it looks like [this refactor](https://github.com/briis/unifiprotect/pull/336/files#diff-c33d3ccf9e61ed4e354cb42e481a9a258c38eb924658d6abe44f8ae0d3432a7cL152-L162) was the cause.

Core uses the [frontend_stream_type](https://github.com/home-assistant/core/blob/dev/homeassistant/components/camera/__init__.py#L423) attribute, which in turn uses the `supported_features` attribute to enable HLS.

### DIFF
--- a/custom_components/unifiprotect/camera.py
+++ b/custom_components/unifiprotect/camera.py
@@ -161,6 +161,11 @@ class UnifiProtectCamera(UnifiProtectEntity, Camera):
         self.async_write_ha_state()
 
     @property
+    def supported_features(self):
+        """Return supported features for this camera."""
+        return self._attr_supported_features
+
+    @property
     def motion_detection_enabled(self):
         """Camera Motion Detection Status."""
         return self._device_data["recording_mode"] and super().available
--- a/.devcontainer/configuration.yaml
+++ b/.devcontainer/configuration.yaml
@@ -1,7 +1,9 @@
 default_config:
 
 stream:
-  ll_hls: True
+  ll_hls: true
+  part_duration: 0.75
+  segment_duration: 2
 
 debugpy:
   start: false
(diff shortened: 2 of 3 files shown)


custom_components/unifiprotect/camera.py
```diff
@@ -160,6 +160,11 @@ def _async_updated_event(self):
         )
         self.async_write_ha_state()
 
+    @property
+    def supported_features(self):
```
bdraco, line 164: 😢 

Camera doesn't support `_attr` yet
https://github.com/home-assistant/core/blob/dev/homeassistant/components/camera/__init__.py#L393

bdraco (approved):

custom_components/unifiprotect/camera.py
AngellusMortis: Oh, is that a newer things some of the other base classes supports?

bdraco: https://developers.home-assistant.io/docs/core/entity/#entity-class-or-instance-attributes

briis:
I just installed the latest stuff, and streaming is working. However I found that the Viewport is no longer a Device, it is just a select entity now. I can't make up my mind if that is a problem, but in reality it is a physical device, so shouldn't it show up under devices?

briis:
Never mind, I was wrong - After a while and a refresh of the Browser it showed up under devices again, sorry.

AngellusMortis:
I bet you it is just because of the need to make the HTTP call to get the list of live views that causes it to take longer to init that device.

PR #349, opened 2021-11-24
Removes references to UnifiOS port configuration

AngellusMortis:
Since UnifiOS is required now, remove references about UnifiOS/Non-UniFiOS devices in port configuration.


AngellusMortis:
@bdraco I think that is all that is needed for that one, right?

bdraco (approved):

PR #350, opened 2021-11-24
Refactors config flow and adds reauth

AngellusMortis:
* Raise `ConfigEntryAuthFailed` on auth failure
* Adds `async_step_reauth` to allow reauth on auth failure
* Refactors `async_step_user` to allow it to be shared between user and reauth steps.
* Updates strings to fix casing of "UniFi"
* Simplifies options flow to only have Disable RTSP and Doorbell Texts

### DIFF
--- a/custom_components/unifiprotect/__init__.py
+++ b/custom_components/unifiprotect/__init__.py
@@ -56,7 +56,7 @@ def _async_import_options_from_data_if_missing(hass: HomeAssistant, entry: Confi
         hass.config_entries.async_update_entry(entry, data=data, options=options)
 
 
-async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
+async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     """Set up the Unifi Protect config entries."""
     _async_import_options_from_data_if_missing(hass, entry)
 
--- a/custom_components/unifiprotect/__init__.py
+++ b/custom_components/unifiprotect/__init__.py
@@ -17,7 +17,7 @@ from homeassistant.const import (
     EVENT_HOMEASSISTANT_STOP,
 )
 from homeassistant.core import HomeAssistant, callback
-from homeassistant.exceptions import ConfigEntryNotReady
+from homeassistant.exceptions import ConfigEntryAuthFailed, ConfigEntryNotReady
 from homeassistant.helpers.aiohttp_client import async_create_clientsession
 import homeassistant.helpers.device_registry as dr
 from pyunifiprotect import NotAuthorized, NvrError, UpvServer
@@ -56,7 +56,7 @@ def _async_import_options_from_data_if_missing(hass: HomeAssistant, entry: Confi
         hass.config_entries.async_update_entry(entry, data=data, options=options)
 
 
-async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
+async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
     """Set up the Unifi Protect config entries."""
     _async_import_options_from_data_if_missing(hass, entry)
 
@@ -74,11 +74,8 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
     try:
         nvr_info = await protectserver.server_information()
-    except NotAuthorized:
-        _LOGGER.error(
-            "Could not Authorize against Unifi Protect. Please reinstall the Integration"
-        )
-        return False
+    except NotAuthorized as err:
+        raise ConfigEntryAuthFailed(err) from err
     except (asyncio.TimeoutError, NvrError, ServerDisconnectedError) as notreadyerror:
         raise ConfigEntryNotReady from notreadyerror
 
--- a/custom_components/unifiprotect/__init__.py
+++ b/custom_components/unifiprotect/__init__.py
@@ -70,7 +70,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     )
 
     _LOGGER.debug("Connect to Unfi Protect")
-    protect_data = UnifiProtectData(hass, protectserver, SCAN_INTERVAL)
+    protect_data = UnifiProtectData(hass, protectserver, SCAN_INTERVAL, entry)
 
     try:
         nvr_info = await protectserver.server_information()
--- a/custom_components/unifiprotect/config_flow.py
+++ b/custom_components/unifiprotect/config_flow.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import logging
+from typing import Optional
 
 from aiohttp import CookieJar
 from homeassistant import config_entries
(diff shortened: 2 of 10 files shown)


custom_components/unifiprotect/__init__.py
```diff
@@ -56,7 +56,7 @@ def _async_import_options_from_data_if_missing(hass: HomeAssistant, entry: Confi
         hass.config_entries.async_update_entry(entry, data=data, options=options)
 
 
-async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
+async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry):
```
bdraco, line 59:
```suggestion
async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
```

This is expected to return True/False

Its only the platforms that don't need it

custom_components/unifiprotect/config_flow.py
```diff
@@ -40,51 +41,93 @@ def async_get_options_flow(config_entry):
...
-            return await self._show_setup_form(user_input)
+        return await self._init_config_entry(user_input=user_input)
 
-        errors = {}
+    def _get_form(self, reauth: bool):
+        form = self._show_setup_form
+        if reauth:
+            form = self._show_reauth_form
+
+        return form
```
bdraco, line 57: If anything is typed, the whole function should be typed.

AngellusMortis: I will go back through the config flow and add typings whereever I can.

custom_components/unifiprotect/config_flow.py
```diff
@@ -96,6 +139,19 @@ async def async_step_user(self, user_input=None):
             },
         )
 
+    async def _show_reauth_form(self, errors=None):
```
bdraco, line 142:
```suggestion
    async def _async_show_reauth_form(self, errors=None):
```

Anything that should run in async context should be prefixed with `async`.  Its a normal function decorate with `@callback` from `homeassistant.core` so Home Assistant internals know to run it in the event loop and not the executor

custom_components/unifiprotect/config_flow.py
```diff
@@ -40,51 +41,93 @@ def async_get_options_flow(config_entry):
...
+            return await form(errors)
+
         try:
-            server_info = await unifiprotect.server_information()
+            server_info = await protect.server_information()
             if server_info["server_version"] < MIN_REQUIRED_PROTECT_V:
                 _LOGGER.debug("UniFi Protect Version not supported")
                 errors["base"] = "protect_version"
-                return await self._show_setup_form(errors)
+                return await form(errors)
```
bdraco, line 111: Usually we build these so everything falls through and shows the form at the end if there are errors.

Simple example: https://github.com/home-assistant/core/blob/dev/homeassistant/components/august/config_flow.py#L123

bdraco: Probably a better example: 
https://github.com/home-assistant/core/blob/dev/homeassistant/components/harmony/config_flow.py#L62

custom_components/unifiprotect/config_flow.py
```diff
@@ -40,51 +41,93 @@ def async_get_options_flow(config_entry):
...
         )
 
+    async def _reload_entry(self, unique_id, user_input):
+        entry = await self.async_set_unique_id(unique_id)
+
+        new_data = entry.data.copy()
+        new_data[CONF_USERNAME] = user_input[CONF_USERNAME]
+        new_data[CONF_PASSWORD] = user_input[CONF_PASSWORD]
+
+        self.hass.config_entries.async_update_entry(entry, data=new_data)
```
bdraco, line 89: Could be written as
```suggestion
        self.hass.config_entries.async_update_entry(entry, data={**entry.data, CONF_USERNAME: user_input[CONF_USERNAME], CONF_PASSWORD: user_input[CONF_PASSWORD]})
```

custom_components/unifiprotect/config_flow.py
```diff
@@ -40,51 +41,93 @@ def async_get_options_flow(config_entry):
...
-            return await self._show_setup_form(errors)
+            return await form(errors)
 
         unique_id = server_info[SERVER_ID]
-        server_name = server_info[SERVER_NAME]
+        if reauth:
+            return await self._reload_entry(unique_id, user_input)
 
-        await self.async_set_unique_id(unique_id)
         self._abort_if_unique_id_configured()
```
bdraco, line 126: The branching is a bit hard to follow above.

Its not immediately clear where unique id is being set before this is called

custom_components/unifiprotect/config_flow.py
bdraco, line 57:
```suggestion
        if reauth:
            return self._show_reauth_form
        return self._show_setup_form
```

custom_components/unifiprotect/config_flow.py
AngellusMortis: I have refactored out some core bits so I am going to back through and split this back into two functions based on your you recommendation about the fall through.

custom_components/unifiprotect/config_flow.py
bdraco: Returning a reference to a function in a config flow isn't a common pattern so this may come back up again in a core review. If we could find a way to avoid it, it would be helpful.

custom_components/unifiprotect/config_flow.py
```diff
@@ -36,79 +49,123 @@ class UnifiProtectFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
...
+
+        return None
+
+    async def async_step_reauth(self, user_input: Dict[str, Any] = None) -> FlowResult:
+        """Perform reauth upon an API authentication error."""
+
+        errors = {}
+        if user_input is not None:
+            entry = self._get_config_entry()
+            if not entry:
```
bdraco, line 130:
```suggestion
            if not (entry := self._get_config_entry()):
```

custom_components/unifiprotect/config_flow.py
```diff
@@ -36,79 +49,123 @@ class UnifiProtectFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
...
-            if server_info["server_version"] < MIN_REQUIRED_PROTECT_V:
+            nvr_data = await protect.server_information()
+            if nvr_data["server_version"] < MIN_REQUIRED_PROTECT_V:
                 _LOGGER.debug("UniFi Protect Version not supported")
                 errors["base"] = "protect_version"
-                return await self._show_setup_form(errors)
-
         except NotAuthorized as ex:
             _LOGGER.debug(ex)
             errors["base"] = "connection_error"
```
bdraco, line 106:
```suggestion
            errors[CONF_PASSWORD] = "connection_error"
```

This is actually the field were the error will show up.   `connection_error` 

We should use the standard `invalid_auth` here so we can use the following in `strings.json`

```
"invalid_auth": "[%key:common::config_flow::error::invalid_auth%]"
```

custom_components/unifiprotect/config_flow.py
```diff
@@ -36,79 +49,123 @@ class UnifiProtectFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
...
 
+        errors = {}
+        nvr_data = None
         try:
-            server_info = await unifiprotect.server_information()
-            if server_info["server_version"] < MIN_REQUIRED_PROTECT_V:
+            nvr_data = await protect.server_information()
+            if nvr_data["server_version"] < MIN_REQUIRED_PROTECT_V:
                 _LOGGER.debug("UniFi Protect Version not supported")
                 errors["base"] = "protect_version"
```
bdraco, line 103: Move code out of the `try` block that isn't expected to fail

bdraco: Probably an `else`?

custom_components/unifiprotect/config_flow.py
```diff
@@ -36,79 +49,123 @@ class UnifiProtectFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
...
-            return await self._show_setup_form(errors)
 
-        unique_id = server_info[SERVER_ID]
-        server_name = server_info[SERVER_NAME]
+        return nvr_data, errors
 
-        await self.async_set_unique_id(unique_id)
-        self._abort_if_unique_id_configured()
+    @callback
+    def _get_config_entry(self) -> Optional[config_entries.ConfigEntry]:
```
bdraco, line 114: You can avoid this function by doing something like this
https://github.com/home-assistant/core/blob/dev/homeassistant/components/ambee/config_flow.py#L74

bdraco: https://github.com/home-assistant/core/blob/dev/homeassistant/config_entries.py#L589

AngellusMortis: The first one was exactly what I was looking for. Thanks.

custom_components/unifiprotect/config_flow.py
AngellusMortis: Decided not to us the walrus operator since it causes a multi-line if (with the other changes to get rid of `_get_config_entry`.

custom_components/unifiprotect/config_flow.py
```diff
@@ -36,79 +38,126 @@ class UnifiProtectFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
...
-                CONF_USERNAME: user_input.get(CONF_USERNAME),
-                CONF_PASSWORD: user_input.get(CONF_PASSWORD),
-            },
-            options={
-                CONF_DISABLE_RTSP: False,
-                CONF_DOORBELL_TEXT: "",
-            },
+        errors = {}
+
+        entry = self.hass.config_entries.async_get_entry(self.context["entry_id"])
```
bdraco, line 103: I think we still need the https://github.com/home-assistant/core/blob/dev/homeassistant/components/ambee/config_flow.py#L74 type logic since I don't think the `self.context` will still have `entity_id` in there if they have to retry because they entered the password wrong

custom_components/unifiprotect/config_flow.py
```diff
@@ -34,6 +34,8 @@
 class UnifiProtectFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
     """Handle a Unifi Protect config flow."""
 
+    entry: Optional[config_entries.ConfigEntry] = None
```
bdraco, line 37: Usually we want to define these in `__init__` to make it clear its not constant between multiple config flows

custom_components/unifiprotect/config_flow.py
```diff
@@ -34,81 +36,140 @@ class UnifiProtectFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
...
+        self.entry = self.hass.config_entries.async_get_entry(self.context["entry_id"])
+        return await self.async_step_reauth_confirm()
+
+    async def async_step_reauth_confirm(
+        self, user_input: dict[str, Any] | None = None
+    ) -> FlowResult:
+
+        errors = {}
+        if self.entry is None:
+            return await self.async_step_user()
```
bdraco, line 115: This is unreachable

AngellusMortis: `async_get_entry` returns an optional `ConfigEntry`. So it _should_ be unreachable, but linters will complain if it is not there.

bdraco: assert its not None instead.   `async_get_entry` doesn't know the reauth flow was started from the config entry itself

custom_components/unifiprotect/config_flow.py
```diff
@@ -34,81 +36,139 @@ class UnifiProtectFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
...
+        assert self.entry is not None
+
+        # prepopulate fields
+        form_data = {**self.entry.data}
+        if user_input is not None:
+            form_data.update(user_input)
+
+            # validate login data
+            nvr_data, errors = await self._async_get_nvr_data(form_data)
+            if nvr_data is not None:
```
bdraco, line 123:
```suggestion
            if not errors:
```

custom_components/unifiprotect/config_flow.py
```diff
@@ -34,81 +36,139 @@ class UnifiProtectFlowHandler(config_entries.ConfigFlow, domain=DOMAIN):
...
+    async def async_step_user(
+        self, user_input: dict[str, Any] | None = None
+    ) -> FlowResult:
+        """Handle a flow initiated by the user."""
+
+        errors = {}
+        if user_input is not None:
+            nvr_data, errors = await self._async_get_nvr_data(user_input)
+
+            if nvr_data is not None:
```
bdraco, line 150:
```suggestion
            if not errors:
```

custom_components/unifiprotect/config_flow.py
bdraco: Should be functionally the same, but shows intent

custom_components/unifiprotect/config_flow.py
bdraco: Should be functionally the same, but shows intent

bdraco (approved):
Testing passed
Screen Shot 2021-11-25 at 11 54 48
Screen Shot 2021-11-25 at 11 54 43
Screen Shot 2021-11-25 at 11 54 32

PR #353, opened 2021-11-25
Adds Verify SSL option to setup flow

AngellusMortis:
Depends on unreleased code in master branch of pyunifiprotect

* Adds Verify SSL to setup flow.
* Adds config entry migration to ensure everyone that migrates keeps the old default of `False`.
* Changes various setup translation keys to use common HA ones

What a SSL failure looks like in HA

image


AngellusMortis:
@bdraco PR is "done" just going to leave it draft though until @briis gets back from the holiday to be able to release a new version of pyunifiprotect.

briis:
I have released 1.0.5 of `pyunifiprotect`, so you can merge now @AngellusMortis

AngellusMortis:
Yep, definitely still want @bdraco to look at it though. It sounds like from him the config flow gets extra eyes for the core PRs.

custom_components/unifiprotect/config_flow.py
```diff
@@ -72,6 +73,7 @@ async def _async_get_nvr_data(
         if CONF_HOST in user_input:
             host = user_input[CONF_HOST]
             port = user_input.get(CONF_PORT, DEFAULT_PORT)
+            verify_ssl = user_input.get(CONF_VERIFY_SSL, True)
```
bdraco, line 76:
```suggestion
            verify_ssl = user_input.get(CONF_VERIFY_SSL, DEFAULT_VERIFY_SSL)
```

Most people won't have a valid ssl so we probably shouldn't check it by default since users won't know what to do to get it to work

 `DEFAULT_VERIFY_SSL` to `False`

custom_components/unifiprotect/config_flow.py
```diff
@@ -162,6 +165,9 @@ async def async_step_user(
                     vol.Required(
                         CONF_PORT, default=user_input.get(CONF_PORT, DEFAULT_PORT)
                     ): int,
+                    vol.Required(
+                        CONF_VERIFY_SSL, default=user_input.get(CONF_VERIFY_SSL, True)
```
bdraco, line 169:
```suggestion
                        CONF_VERIFY_SSL, default=user_input.get(CONF_VERIFY_SSL, DEFAULT_VERIFY_SSL)
```


Most people won't have a valid ssl so we probably shouldn't check it by default since users won't know what to do to get it to work

 `DEFAULT_VERIFY_SSL` to `False`

custom_components/unifiprotect/config_flow.py
AngellusMortis: I figured you were going to say that. I am really torn on that, but I cannot argue with you that most people will have invalid certs, but it is still a "bad practice" to default to false and the Devops in me hates it.

bdraco: Since installing a valid ssl requires user interaction with the device, we should hope they are also proficient enough to check the box 🤣

bdraco (approved):

Issue #341, opened 2021-11-19
Binary Sensor Doorbell not triggered


BoisMaison:
### Home Assistant Version?

2021.10.6

### UniFi Protect Version?

1.19.2

### UniFi Protect HA Integration version?

0.9.1

### What happened?

When the doorbell is pressed, the binary sensor (doorbell) is not triggered, meaning it cannot be used in home assistant. All other triggers (motion detection) work flawless.

AngellusMortis:
I am not able to reproduce and I just got a package this morning. I am on the latest 0.10 beta, but that should not matter since the core entities have not been changed for 0.10.

![image](https://user-images.githubusercontent.com/490848/142672685-348e4c72-c1ba-464c-b873-8004e211332e.png)

There are binary sensors two sensors for cameras with a chime (like the G4 Doorbell). One is for motion and one is for the chime.

Make sure you are using the right one. You want to make sure you use the one with the doorbell icon, not the one with the walking icon.

![image](https://user-images.githubusercontent.com/490848/142672746-cceac644-a991-4577-b511-636dd8c1caff.png)

briis:
You could try and use the event that is also fired when the doorbell is pressed. See [this example](https://github.com/briis/unifiprotect#send-a-notification-when-the-doorbell-is-pressed) in the documentation.

@AngellusMortis I have seen this before for some users. It might be a timing issue. So maybe we should see if we could let the sensor be on for 2-3 seconds before reverting to off.

BoisMaison:
Maybe I had to try it before, but after a full reboot and reinstallation of the UniFi doorbell itself, it works like a charm now. Might be useful to for other who face this (still strange) issue.

briis:
All that shouldn't be necessary, but sometimes with Ubiquity HW and SW it just needs some love (and reboots)

KasperHoldum:
I experience the same thing. Most of the time the doorbell event is fired, but sometimes it is not. Which makes the doorbell unreliable.

AngellusMortis:
I added code to the 0.11 to handle this case better. Are able to install it and test it out?

BoisMaison:
So far the integration works flawless, I did not need another reset anymore.

AngellusMortis:
I am going to go ahead and close this one as confirms fixed in 0.11 then. Thanks for checking and replying.

Issue #317, opened 2021-10-27
Motion Sensor entity does not survive Device becoming unavailable


jb123567:
If a camera becomes unavailable, for example due to a firmware update or a network update the motion sensor entity will no longer update and HA needs to be restarted.

AngellusMortis:
This should be resolved in the 0.11 beta. Are you able to install it and test it out?

jb123567:
Already installed, will let you know if the error occurs again!

AngellusMortis:
It sounds like the issue is fixed since we have not heard back. I am going to close this for now and if you have any more issues, let us know!

Issue #297, opened 2021-09-10
Websocket connection not active error in logs


copart:
HASS Version | core-2021.9.3 (in docker)
HACS: 1.15.2
UniFi Protect Integration: 0.9.0

[pyunifiprotect.unifi_protect_server] Unifi OS: Websocket connection not active, failing back to polling

gokou340:
I'm also getting the same thing. What version of Protect are you on? I am on 1.19.1 Beta 4

briis:
Hi,
This is not an issue with this integration as far as we can see. It is most likely related to a general network issue. Reloading the UniFi Protect will normally fix the issue, should it occur.
The logging of this, has only just been implemented in the latest release, so it might have been there earlier, just without the log entry. We are looking at not posting the error, unless it continues to occur.

copart:
Protect version 1.19.0 

Anyway to confirm what IP address the integration is using to access the server? Hoping its local and not Internet IP.

briis:
Not sure I understand the question. The Integration uses the IP Address you key in the Config Dialog when setting it up, and that would most often be a local IP address.

Veldkornet:
I had a look at my HomeAssistant log today only to find 50,000+ lines of this
```
2021-09-22 14:33:05 WARNING (MainThread) [pyunifiprotect.unifi_protect_server] Unifi OS: Websocket connection not active, failing back to polling
2021-09-22 14:33:07 WARNING (MainThread) [pyunifiprotect.unifi_protect_server] Unifi OS: Websocket connection not active, failing back to polling
2021-09-22 14:33:09 WARNING (MainThread) [pyunifiprotect.unifi_protect_server] Unifi OS: Websocket connection not active, failing back to polling
2021-09-22 14:33:11 WARNING (MainThread) [pyunifiprotect.unifi_protect_server] Unifi OS: Websocket connection not active, failing back to polling
2021-09-22 14:33:13 WARNING (MainThread) [pyunifiprotect.unifi_protect_server] Unifi OS: Websocket connection not active, failing back to polling
2021-09-22 14:33:15 WARNING (MainThread) [pyunifiprotect.unifi_protect_server] Unifi OS: Websocket connection not active, failing back to polling
2021-09-22 14:33:17 WARNING (MainThread) [pyunifiprotect.unifi_protect_server] Unifi OS: Websocket connection not active, failing back to polling
[...]

Issue #255, opened 2021-07-01
No more binary_sensor updates after NvrError


andrewkroh:
My Unifi controller rebooted and I think this integration may not have fully recovered after that occurred. There were a few connectivity related exceptions (`ClientConnectorError`) followed by one `NvrError`. That last exception might have been while the controller was still coming up.

```
2021-06-30 19:45:43 ERROR (MainThread) [custom_components.unifiprotect.data] Error while updating
Traceback (most recent call last):
  File "/config/custom_components/unifiprotect/data.py", line 48, in async_refresh
    await self._protectserver.update(
  File "/usr/local/lib/python3.8/site-packages/pyunifiprotect/unifi_protect_server.py", line 123, in update
    await self._get_device_list(not self.ws_connection)
  File "/usr/local/lib/python3.8/site-packages/pyunifiprotect/unifi_protect_server.py", line 306, in _get_device_list
    raise NvrError(
pyunifiprotect.unifi_protect_server.NvrError: Fetching Camera List failed: 404 - Reason: Not Found
```

Version: v0.8.8
HA Version: 2021.5.5
Attachment: [home-assistant-logs.csv](https://github.com/briis/unifiprotect/files/6748390/home-assistant-logs.csv)

After the NvrError at 19:45, home assistant stopped getting any `sensor` and `binary_sensor` updates related to Unifi Protect. Only the `camera` entities were updating (every 5m). I was logging events with the [Elasticsearch integration](https://github.com/legrego/homeassistant-elasticsearch) so I took a look at the data:

<img width="1072" alt="Screen Shot 2021-07-01 at 8 35 49 AM" src="https://user-images.githubusercontent.com/4565752/124125934-251ba000-da48-11eb-9008-4ef6537f92b3.png">

I thought maybe the code was missing some exception handling around this line to trigger a retry (but I'm not at all familiar with writing home assistant integrations).
https://github.com/briis/unifiprotect/blob/631cfd64b5c9cb1a72c605a5b0c404d7d98b1aec/custom_components/unifiprotect/__init__.py#L99

briis:
Thanks for the detailed log, looks good, I need to look at that Integration 😁

> await protect_data.async_setup() 

If you look at the next two lines after the above, they first check if we got data returned, and if not the line `raise ConfigEntryNotReady` will then retry getting data, or eventually give up. If have tested this many times, by pulling the plug on my Protect Server, and my entity always recovers after that. I do get entries in the log files, but I do not see it crash completely.

[...]

PR #354, opened 2021-11-25
Migration to using `ProtectApiClient`

### LINKED ISSUE #337: Camera state always shows Recording even though nothing is being recorded
### Home Assistant Version?

2021.11.2

### UniFi Protect Version?

1.19.2

### UniFi Protect HA Integration version?

0.9.1

### What happened?

Camera is set to record only during a motion event. 
The camera state in HA shows Recording even though there is no motion event and nothing is being recorded on the Cloudkey gen2+.

Expected: 
If no motion event, the video stream is not being recorded and therefore the camera state in HA should show Idle.

### Relevant log output

_No response_

AngellusMortis:
* `NEW`: **BREAKING CHANGE** Adds all of the possible enabled UFP Camera channels as different camera entities; only the highest resolution secure (RTSPS) one is enabled by default. This will change the unique ID of your default Camera so your old Camera entities will become orphaned. If you need RTSP camera entities, you can disable one of the given insecure camera entities.

Note: this means there are now up to 6 camera entities with only a single one enabled by default.
image

* `CHANGE`: **BREAKING CHANGE** The internal name of the Privacy Zone controlled by the "Privacy Mode" switch has been changed. Make sure you turn off all of your privacy mode switches before upgrading.

* `CHANGE`: **BREAKING CHANGE** WDR `number` entity has been removed from Cameras that have HDR. This is inline with changes made to Protect as you can no longer control WDR for cameras with HDR.

* `CHANGE`: **BREAKING CHANGE** `event_length` attribute has been removed from the motion and door binary sensors. The value was previously calculated in memory and not reliable between restarts.

* `CHANGE`: **BREAKING CHANGE** the `event_object` attribute for binary motion sensors has changed the value for no object detected from "None Identified" (string) to "None" (NoneType/null)

* `CHANGE`: Migrates `UpvServer` to new `ProtectApiClient` from `pyunifiprotect`.
    * This should lead to a number of behind-the-scenes reliability improvements.
      * Should fix/close the following issues: #248, #255, #297, #317, #341, and #360 (TODO: Verify after we make a beta version)

* `CHANGE`: The state of the camera entities now reflects on whether the camera is actually recording. If you set your Recording Mode to "Detections", your camera will switch back and forth between "Idle" and "Recording" based on if the camera is actually recording.
  * Closes #337

* `CHANGE`: Configuration URLs for UFP devices will now take you directly to the device in the UFP Web UI.

* `CHANGE`: Default names for all entities have been updated from `entity_name device_name` to `device_name entity_name` to match how Home Assistant expects them in 2021.11+

* `NEW`: Added the following attributes to Camera entity: `width`, `height`, `fps`, `bitrate` and `channel_id`

* `NEW`: Added status light switch for Flood Light devices

* `NEW`: Added "On Motion - When Dark" option for Flood Light Lighting switch

* `NEW`: Added "Auto-Shutoff Timer" number entity for Flood Lights

* `NEW`: Added "Motion Sensitivity" number entity for Flood Lights

* `NEW`: Added "Chime Duration" number entity for Doorbells


briis:
Cool. Will start testing, once I am back home.

AngellusMortis:
Okay. This should be everything. I have gone through and done a bunch of basic testing and everything looks good. I will do another round of more in depth testing again tomorrow and may push up some more changes then.

briis:
Really nice, I can't wait to come home and give this a test.

AngellusMortis:
Okay. I am done with testing everything looks good for me. Pushed a bunch of fixes upstream as well as some here. I also added a bunch of entities that will allow attributes / services to be removed.

I added a deprecation section in the WIP changelog for everything that will be removing in the core merge as per @bdraco's list. Please make sure you review it and see if there is anything on it I missed and/or we do not have another entity to cover for it.

briis:
Not sure if it is ready to test yet, but I pulled this branch, and I keep getting this error:

```log
2021-11-28 08:07:15 ERROR (MainThread) [pyunifiprotect.api] Exception while running subscription handler
  File "/usr/local/python/lib/python3.9/site-packages/pyunifiprotect/api.py", line 548, in _process_ws_message
    sub(processed_message)
  File "/config/custom_components/unifiprotect/data.py", line 93, in _async_process_ws_message
    self.async_signal_device_id_update(message.new_obj.id)
  File "/config/custom_components/unifiprotect/data.py", line 144, in async_signal_device_id_update
    update_callback()
  File "/config/custom_components/unifiprotect/entity.py", line 77, in _async_updated_event
    self.async_write_ha_state()
  File "/usr/local/python/lib/python3.9/site-packages/homeassistant/helpers/entity.py", line 481, in async_write_ha_state
    self._async_write_ha_state()
  File "/usr/local/python/lib/python3.9/site-packages/homeassistant/helpers/entity.py", line 517, in _async_write_ha_state
    extra_state_attributes = self.extra_state_attributes
  File "/config/custom_components/unifiprotect/camera.py", line 246, in extra_state_attributes
    ATTR_PRIVACY_MODE: self.device.is_privacy_on,
AttributeError: 'Camera' object has no attribute 'is_privacy_on'
````
Which then leads to all sorts of other error, with recording state missing, cameras that won't stream etc.

**Comment by AngellusMortis** [comment]
Looks like you are missing the required pyunifprotect package. You have to manually install it from the listed PR since that PR is not merged yet.

**Comment by briis** [comment]
Bummer - Forgot that. Now it's working. Will do some testing....

**Comment by AngellusMortis** [comment]
After talking with @bdraco, he definitely likes the idea of just adding _all of the camera_ and then only enabling the "default" one. That then gives the users the freedom to use which everone they actually want to use.

**Comment by briis** [comment]
Sounds like a good plan. I guess we underestimated how many were dependent on the RTSP stream.

**Comment by AngellusMortis** [comment]
For sure. As long as we are providing the default as one that will work going forward, we should be good if they ever decide to remove RTSP in the future.

**Comment by AngellusMortis** [comment]
Found a number of typos that are breaking changes if you have already installed this version into HA. You will probably get some orphaned entitles.

**Comment by bdraco** [comment]
` If you need RTSP camera entities, you can disable one of the given insecure camera entities.`. Should that saw `enable`?

File: custom_components/unifiprotect/data.py [hunk 1]
```diff
@@ -28,38 +24,41 @@ class UnifiProtectData:
...
-        if not self.data:
-            return
 
-        for device_id, device_data in self.data.items():
-            device_type = device_data.get("type")
-            if device_type and device_type in device_types:
-                yield UnifiProtectDevice(device_id, device_type, device_data)
+        attrs = [f"{m.value}s" for m in device_types]
+
+        for attr in attrs:
```
bdraco, line 51: Can we avoid iterating twice ?

custom_components/unifiprotect/data.py
```diff
@@ -93,11 +88,27 @@ async def async_refresh(self, *_, force_camera_update=False):
...
-        for device_id, data in updates.items():
-            self.data[device_id] = data
-            self.async_signal_device_id_update(device_id)
+
+        # Websocket connected, use data from it
+        if updates is None:
+            return
+
+        attrs = [f"{m.value}s" for m in DEVICES_WITH_ENTITIES]
+        for attr in attrs:
```
bdraco, line 106: Can we avoid iterating twice ?

custom_components/unifiprotect/entity.py
```diff
@@ -5,45 +5,45 @@
...
-        self._attr_name = self._device_name
-        self._mac = self._device_data["mac"]
-        self._firmware_version = self._device_data["firmware_version"]
-        self._server_id = server_info["server_id"]
-        self._server_ip = server_info["server_ip"]
-        self._device_type = self._device_data["type"]
-        self._model = self._device_data["model"]
+        self.protect: ProtectApiClient = protect
+        self.protect_data: UnifiProtectData = protect_data
+        self.device: ProtectAdoptableDeviceModel = device
```
bdraco, line 34: This is getting set in the child class sometimes.  Can we avoid that?

custom_components/unifiprotect/light.py
```diff
@@ -71,56 +72,64 @@ def hass_to_unifi_brightness(value):
...
         """Initialize an Unifi light."""
-        super().__init__(upv_object, protect_data, server_info, light_id, None)
-        self._name = self._device_data["name"]
+        super().__init__(protect, protect_data, device, None)
+        self.device: Light = device
+        self._attr_name = self.device.name
         self._attr_icon = "mdi:spotlight-beam"
         self._attr_supported_features = SUPPORT_BRIGHTNESS
-        self._attr_is_on = self._device_data["is_on"] == ON_STATE
+        self._async_update_device_from_protect()
```
bdraco, line 82: TODO: check this to make sure it doesn't write the state twice at startup

custom_components/unifiprotect/select.py
```diff
@@ -207,27 +238,46 @@ class UnifiProtectSelects(UnifiProtectEntity, SelectEntity):
...
-        unifi_value = self._device_data[self._data_key]
-        if self.entity_description.key == _KEY_DOORBELL_TEXT:
-            return unifi_value
+        unifi_value = get_nested_attr(self.device, self._data_key)
+
+        if isinstance(unifi_value, Enum):
+            unifi_value = unifi_value.value
+
+        if isinstance(unifi_value, Liveview):
+            unifi_value = unifi_value.id
```
bdraco, line 265: Can this be an `if/elif` ?

custom_components/unifiprotect/select.py
```diff
@@ -237,22 +287,36 @@ async def async_select_option(self, option: str) -> None:
...
-            )
-            _LOGGER.debug("Changed Doorbell LCD Text to: %s", option)
-            return
+        if isinstance(self.device, Camera):
+            if self.entity_description.key == _KEY_DOORBELL_TEXT:
+                if unifi_value.startswith(DoorbellMessageType.CUSTOM_MESSAGE.value):
+                    await self.device.set_lcd_text(
+                        DoorbellMessageType.CUSTOM_MESSAGE, text=option
+                    )
+                elif unifi_value == "":
```
bdraco, line 307: Maybe make `""` a constant so its more clear whats going on here?

custom_components/unifiprotect/select.py
```diff
@@ -207,27 +238,46 @@ class UnifiProtectSelects(UnifiProtectEntity, SelectEntity):
...
+            # a bit of extra to allow On Motion Always/Dark
+            if (
+                self.device.light_mode_settings.mode == LightModeType.MOTION
+                and self.device.light_mode_settings.enable_at
+                == LightModeEnableType.DARK
+            ):
+                unifi_value = f"{LightModeType.MOTION.value}Dark"
+        elif self.entity_description.key == _KEY_DOORBELL_TEXT:
+            if unifi_value is None:
+                unifi_value = ""
```
bdraco, line 278: Maybe make `""` a constant so its more clear whats going on here?

custom_components/unifiprotect/switch.py
```diff
@@ -118,67 +126,68 @@ class UnifiProtectSwitch(UnifiProtectEntity, SwitchEntity):
...
         description: UnifiProtectSwitchEntityDescription,
     ):
         """Initialize an Unifi Protect Switch."""
-        super().__init__(upv_object, protect_data, server_info, device_id, description)
-        self._attr_name = f"{self.entity_description.name} {self._device_data['name']}"
+        super().__init__(protect, protect_data, device, description)
+        self.device: Camera | Light = device
+        self._attr_name = f"{self.device.name} {self.entity_description.name}"
         self._switch_type = self.entity_description.key
+        if isinstance(self.device, Camera):
```
bdraco, line 139: Could reverse this an return early + outdent below

briis:
The `last_tripped_time` attribute has changed from a Datetime to an epoch. Should be mentioned in the Release Notes.

custom_components/unifiprotect/switch.py
```diff
@@ -118,67 +126,68 @@ class UnifiProtectSwitch(UnifiProtectEntity, SwitchEntity):
...
-        elif self._switch_type == _KEY_PRIVACY_MODE:
-            _LOGGER.debug("Turning Privacy Mode on for %s", self._device_data["name"])
-            self._device_data["save_mic_level"] = self._device_data["mic_volume"]
-            self._device_data["save_rec_mode"] = self._device_data["recording_mode"]
-            await self.upv_object.set_privacy_mode(
-                self._device_id, True, 0, TYPE_RECORD_NEVER
-            )
-        else:
+
+        if isinstance(self.device, Camera):
```
bdraco, line 158: Could reverse this and move the move up the outer block and then return to save some indent.

custom_components/unifiprotect/camera.py
```diff
@@ -178,123 +266,120 @@ def brand(self):
...
+            ATTR_IS_DARK: self.device.is_dark,
+            ATTR_MIC_SENSITIVITY: self.device.mic_volume,
+            ATTR_PRIVACY_MODE: self.device.is_privacy_on,
+            ATTR_WDR_VALUE: self.device.isp_settings.wdr,
+            ATTR_ZOOM_POSITION: self.device.isp_settings.zoom_position,
+            ATTR_WIDTH: self.channel.width,
+            ATTR_HEIGHT: self.channel.height,
+            ATTR_FPS: self.channel.fps,
+            ATTR_BITRATE: self.channel.bitrate,
+            ATTR_CHANNEL_ID: self.channel.id,
```
bdraco, line 295: Some of these are duplicated with sensors and should be dropped here

custom_components/unifiprotect/camera.py
```diff
@@ -178,123 +266,120 @@ def brand(self):
...
-            ATTR_UP_SINCE: self._device_data["up_since"],
-            ATTR_CAMERA_ID: self._device_id,
-            ATTR_CHIME_ENABLED: chime_enabled,
-            ATTR_CHIME_DURATION: self._device_data["chime_duration"],
-            ATTR_IS_DARK: self._device_data["is_dark"],
-            ATTR_MIC_SENSITIVITY: self._device_data["mic_volume"],
-            ATTR_PRIVACY_MODE: self._device_data["privacy_on"],
-            ATTR_WDR_VALUE: self._device_data["wdr"],
-            ATTR_ZOOM_POSITION: self._device_data["zoom_position"],
+            ATTR_UP_SINCE: to_js_time(self.device.up_since),
```
bdraco, line 282: All datetime attributes are required to be in utc isoformat.  Its prefered to make them sensors though since sensors now require utc datetime objects.

bdraco:
I'll run though this again after some sleep.  I'm too wiped out to finish tonight.

bdraco:
> The `last_tripped_time` attribute has changed from a Datetime to an epoch. Should be mentioned in the Release Notes.

All datetime attributes are required to be in utc isoformat.   note that sensors now require utc datetime objects (previously they were required to be utc isoformat)

AngellusMortis:
> ` If you need RTSP camera entities, you can disable one of the given insecure camera entities.`. Should that saw `enable`?

I already fixed that typo in the code.

AngellusMortis:
> > The `last_tripped_time` attribute has changed from a Datetime to an epoch. Should be mentioned in the Release Notes.
> 
> All datetime attributes are required to be in utc isoformat. note that sensors now require utc datetime objects (previously they were required to be utc isoformat)

Shit. Good catch. I thought they were already the Epoch ones. That should be easy to fix.

custom_components/unifiprotect/select.py
AngellusMortis: That was on purpose. Originally. An earlier iteration of this had it so the value could be an Enum and still hit one of the follow ifs. Does not look like it is needed now though.

custom_components/unifiprotect/light.py
AngellusMortis: It is not. That is why I made `_async_update_device_from_protect` (as opposed to the existing `_async_updated_event`)

custom_components/unifiprotect/camera.py
AngellusMortis: I left them in here for now to prevent additional breaking changes. If you look in the change log, many of these attributes are set as deprecated (I think for the Camera, _all_ of them except the new ones I added: width, height, fps, bitrate and channel ID)

AngellusMortis:
Make sure you guys both review the changelog. Especially the deprecations section. That is everything I was thinking that we were remove for when we merge to core based on @bdraco's list of TODOs. 

`last_tripped_time` is one of them. I was thinking that it was redundant since HA would have a record of the last time the second was active.

custom_components/unifiprotect/entity.py
AngellusMortis: I did that on purpose to reset the type of the object. For when it is a more specific child class (ex: in `camera.py` it is _always_ a camera object). You mentioned now having the class attributes outside of the init before. 

Is there a better way to do that?

CHANGELOG.md
```diff
@@ -2,11 +2,79 @@
...
+* `unifiprotect_doorbell` -- use a State Changed event on "Doorbell" binary sensor on the device instead
+* `unifiprotect_motion` -- use a State Changed event on the "Motion" binary sensor on the device instead
+
+All of following attributes should be duplicated data that can be gotten from other devices/entities and as such, they will be removed in the next version.
+
+* `device_model` will be removed from all entities
+* `last_tripped_time` will be removed from binary sensor entities
+* `up_since` will be removed from camera and light entities
+* `enabled_at` will be removed from light entities
+* `camera_id`, `chime_duration`, `is_dark`, `mic_sensitivity`, `privacy_mode`, `wdr_value`, and `zoom_position`  will be removed from camera entities
```
AngellusMortis, line 35: Make sure you guys both review the changelog. Especially the deprecations section. That is everything I was thinking that we were remove for when we merge to core based on @bdraco's list of TODOs.

last_tripped_time is one of them. I was thinking that it was redundant since HA would have a record of the last time the second was active.

briis: The services were initially created to make it possible to set some values from an automation, which was not really possible from any of the entities. With the addition of the Select and Number entity plus a few new Switch entities, I agree that these can be removed, as all of the Services now can be executed from an automation using the built-in entity services.

Of the attributes, it is only the last_tripped_time I am not sure about can be removed, as this was specifically requested by users many releases ago. The last_changed time in HA, just indicates when it changed state the last time, and not when it came on - as far as I know?

My five cents.

AngellusMortis: > The last_changed time in HA, just indicates when it changed state the last time, and not when it came on - as far as I know?

That is what I was unsure about as well. That is why I added it to the list, so we could talk about it. It is a binary sensor, so I would assume the last changed time _was_ the last time it was tripped.

AngellusMortis: I also just added the "Motion Recording" sensors (cameras) and "Light Turn On" sensors (lights) to this list as well since the selects that exist are even better.

briis: > so I would assume the last changed time was the last time it was tripped.

I will do a quick test on that.

briis: Test completed. The last_changed shows when the state was last changed. So going from on to off, resets the last_changed value. Therefore I think we should keep the last_tripped_time.

AngellusMortis: From my testing, they almost lined up perfectly. 

image

AngellusMortis: Unfortunately, as the current code for how it works stands, that is how `last_tripped_time` works now as well. The `last_motion` attribute that comes from UniFi Protect is constantly updated until the motion event ends. So `last_motion` == `last_changed` within a reasonably degree of accurately.

AngellusMortis: Added logging to test/verify this:

```
2021-11-29 14:22:25 DEBUG (MainThread) [pyunifiprotect.api] Skipping update since websocket is active
2021-11-29 14:22:27 DEBUG (MainThread) [custom_components.unifiprotect.data] {'last_motion': datetime.datetime(2021, 11, 29, 14, 22, 27, 235000, tzinfo=datetime.timezone.utc), 'is_motion_detected': False, 'is_recording': True}
2021-11-29 14:22:27 DEBUG (MainThread) [custom_components.unifiprotect.data] {'last_motion': datetime.datetime(2021, 11, 29, 14, 22, 35, 568000, tzinfo=datetime.timezone.utc)}
2021-11-29 14:22:27 DEBUG (MainThread) [pyunifiprotect.api] Skipping update since websocket is active
2021-11-29 14:22:29 DEBUG (MainThread) [pyunifiprotect.api] Skipping update since websocket is active
2021-11-29 14:22:31 DEBUG (MainThread) [custom_components.unifiprotect.data] {'last_motion': datetime.datetime(2021, 11, 29, 14, 22, 31, 669000, tzinfo=datetime.timezone.utc), 'is_motion_detected': False, 'is_recording': True}
2021-11-29 14:22:31 DEBUG (MainThread) [pyunifiprotect.api] Skipping update since websocket is active
2021-11-29 14:22:31 DEBUG (MainThread) [custom_components.unifiprotect.data] {'last_motion': datetime.datetime(2021, 11, 29, 14, 22, 39, 668000, tzinfo=datetime.timezone.utc)}
```

When `is_motion_detected: False` is _not_ passed, that is a message to update the `last_motion` timestamp because motion is ongoing.

briis: OK, I get this. The still remaining issue is that if people use this attribute in automations, it is much easier from a user perspective to check for `last_tripped_time` than to figure out how to get the last updated. So we should at least document that in the release notes.

AngellusMortis: I am updating the attributes in the changelog for where to get the data from instead. 

* `device_model`: Might be more for @bdraco, but how do I get data about the entity's device? The model is on the device itself, but how do you get that for an automation? I now the devices are more of an abstract, but that data is still there in the `entity_description`
* `last_tripped_time` -> `last_changed`
* `up_since` -> Is there a way to get how long it has been since an entity was unavailable? Otherwise, should we keep this one?
* `camera_id`: Looking at it, we do not actually store the device ID on the device/`entity_description`, so maybe we should keep this one? Still deprecate it and move it to `device_id` on _on_ entities?

briis: Many of these attributes were added in the early days of the Integration, and were more used by me to debug stuff, than for actual practical use. So in my book, only last_tripped_time needs to be documented. The rest can just be deprecated. I will be surprised if anyone ask for them back.

AngellusMortis: Updates the defaults on each of them and added a new "Uptime" sensor to replace the `up_since` attribute.

custom_components/unifiprotect/binary_sensor.py
```diff
@@ -180,64 +240,115 @@ def _async_fire_events(self):
...
         ):
-            detected_object = self._device_data["event_object"][0]
+            score = self.device.last_smart_detect_event.score
+            detected_object = self.device.last_smart_detect_event.smart_detect_types[0]
             _LOGGER.debug(
-                "OBJECTS: %s on %s", self._device_data["event_object"], self._attr_name
+                "OBJECTS: %s on %s",
+                self.device.last_smart_detect_event.smart_detect_types,
+                self._attr_name,
             )
```
bdraco, line 339: This is existing, but side effects aren't allowed in `property` methods

custom_components/unifiprotect/camera.py
```diff
@@ -139,26 +194,59 @@ class UnifiProtectCamera(UnifiProtectEntity, Camera):
...
-        self._attr_available = (
-            self._device_data["online"] and self.protect_data.last_update_success
-        )
-        self.async_write_ha_state()
+    def _async_update_device_from_protect(self):
+        super()._async_update_device_from_protect()
+        for channel in self.device.channels:
+            if channel.id == self.channel.id:
+                self.channel = channel
+                break
```
bdraco, line 248: Can we avoid a linear search here?

custom_components/unifiprotect/entity.py
bdraco: Always set `self.device` in the child class before calling `super()`

custom_components/unifiprotect/binary_sensor.py
AngellusMortis: I can refactor that to be cleaner.

custom_components/unifiprotect/sensor.py
```diff
@@ -141,27 +151,31 @@ class UnifiProtectSensor(UnifiProtectEntity, SensorEntity):
...
+        self._attr_name = f"{self.device.name} {self.entity_description.name}"
 
     @property
     def native_value(self):
         """Return the state of the sensor."""
-        return self._device_data[self.entity_description.ufp_value]
+        value = get_nested_attr(self.device, self.entity_description.ufp_value)
+
+        if isinstance(value, datetime):
+            value = value.replace(microsecond=0).isoformat()
```
bdraco, line 169:
```suggestion
            return value.replace(microsecond=0).isoformat()
```

custom_components/unifiprotect/switch.py
```diff
@@ -118,67 +126,73 @@ class UnifiProtectSwitch(UnifiProtectEntity, SwitchEntity):
...
+                _LOGGER.debug("Turning off HDR mode")
+                await self.device.set_hdr(False)
+            elif self._switch_type == _KEY_HIGH_FPS:
+                _LOGGER.debug("Turning off High FPS mode")
+                await self.device.set_video_mode(VideoMode.DEFAULT)
+            elif self._switch_type == _KEY_PRIVACY_MODE:
+                _LOGGER.debug("Turning Privacy Mode off for %s", self.device.name)
+                await self.device.set_privacy(
+                    False, self._previous_mic_level, self._previous_record_mode
+                )
```
bdraco, line 194:
```suggestion
                )
                return
```

custom_components/unifiprotect/switch.py
```diff
@@ -118,67 +126,73 @@ class UnifiProtectSwitch(UnifiProtectEntity, SwitchEntity):
...
     async def async_turn_off(self, **kwargs):
         """Turn the device off."""
+
+        if isinstance(self.device, Camera):
+            if self._switch_type == _KEY_HDR_MODE:
+                _LOGGER.debug("Turning off HDR mode")
+                await self.device.set_hdr(False)
+            elif self._switch_type == _KEY_HIGH_FPS:
+                _LOGGER.debug("Turning off High FPS mode")
+                await self.device.set_video_mode(VideoMode.DEFAULT)
```
bdraco, line 189:
```suggestion
                await self.device.set_video_mode(VideoMode.DEFAULT)
                return
```

custom_components/unifiprotect/switch.py
```diff
@@ -118,67 +126,73 @@ class UnifiProtectSwitch(UnifiProtectEntity, SwitchEntity):
...
+            self._previous_record_mode = self.device.recording_settings.mode
+            await self.device.set_privacy(True, 0, RecordingMode.NEVER)
 
     async def async_turn_off(self, **kwargs):
         """Turn the device off."""
+
+        if isinstance(self.device, Camera):
+            if self._switch_type == _KEY_HDR_MODE:
+                _LOGGER.debug("Turning off HDR mode")
+                await self.device.set_hdr(False)
```
bdraco, line 186:
```suggestion
                await self.device.set_hdr(False)
                return
```

custom_components/unifiprotect/switch.py
```diff
@@ -118,67 +126,73 @@ class UnifiProtectSwitch(UnifiProtectEntity, SwitchEntity):
...
+            elif self._switch_type == _KEY_HIGH_FPS:
```
bdraco, line 187:
```suggestion
            if self._switch_type == _KEY_HIGH_FPS:
```

custom_components/unifiprotect/switch.py
```diff
@@ -118,67 +126,73 @@ class UnifiProtectSwitch(UnifiProtectEntity, SwitchEntity):
...
+                _LOGGER.debug("Turning off High FPS mode")
+                await self.device.set_video_mode(VideoMode.DEFAULT)
+            elif self._switch_type == _KEY_PRIVACY_MODE:
```
bdraco, line 190:
```suggestion
            if self._switch_type == _KEY_PRIVACY_MODE:
```

custom_components/unifiprotect/camera.py
```diff
@@ -139,26 +194,59 @@ class UnifiProtectCamera(UnifiProtectEntity, Camera):
...
         self._last_image = None
+        self._async_set_stream_source()
+        if self._secure:
+            self._attr_unique_id = (
+                f"{self.device.id}_{self.device.mac}_{self.channel.id}"
+            )
+        else:
+            self._attr_unique_id = (
+                f"{self.device.id}_{self.device.mac}_{self.channel.id}_insecure"
+            )
```
bdraco, line 221: Do we need both `device.id` and `device.mac`?  mac should be enough?

custom_components/unifiprotect/switch.py
```diff
@@ -118,67 +126,73 @@ class UnifiProtectSwitch(UnifiProtectEntity, SwitchEntity):
...
+            _LOGGER.debug("Changing Status Light to On")
+            await self.device.set_status_light(True)
+
+        if isinstance(self.device, Light):
+            return
+
         if self._switch_type == _KEY_HDR_MODE:
             _LOGGER.debug("Turning on HDR mode")
-            await self.upv_object.set_camera_hdr_mode(self._device_id, True)
+            await self.device.set_hdr(True)
```
bdraco, line 170:
```suggestion
            await self.device.set_hdr(True)
            return
```

custom_components/unifiprotect/switch.py
```diff
@@ -118,67 +126,73 @@ class UnifiProtectSwitch(UnifiProtectEntity, SwitchEntity):
...
         elif self._switch_type == _KEY_HIGH_FPS:
```
bdraco, line 171:
```suggestion
        if self._switch_type == _KEY_HIGH_FPS:
```

custom_components/unifiprotect/switch.py
```diff
@@ -118,67 +126,73 @@ class UnifiProtectSwitch(UnifiProtectEntity, SwitchEntity):
...
             _LOGGER.debug("Turning on High FPS mode")
-            await self.upv_object.set_camera_video_mode_highfps(self._device_id, True)
+            await self.device.set_video_mode(VideoMode.HIGH_FPS)
```
bdraco, line 173:
```suggestion
            await self.device.set_video_mode(VideoMode.HIGH_FPS)
            return
```

custom_components/unifiprotect/switch.py
```diff
@@ -118,67 +126,73 @@ class UnifiProtectSwitch(UnifiProtectEntity, SwitchEntity):
...
         elif self._switch_type == _KEY_PRIVACY_MODE:
```
bdraco, line 174:
```suggestion
        if self._switch_type == _KEY_PRIVACY_MODE:
```

custom_components/unifiprotect/switch.py
```diff
@@ -118,67 +126,73 @@ class UnifiProtectSwitch(UnifiProtectEntity, SwitchEntity):
...
-        else:
-            _LOGGER.debug("Changing Status Light to On")
-            await self.upv_object.set_device_status_light(
-                self._device_id, True, self._device_type
-            )
-        await self.protect_data.async_refresh(force_camera_update=True)
+            _LOGGER.debug("Turning Privacy Mode on for %s", self.device.name)
+            self._previous_mic_level = self.device.mic_volume
+            self._previous_record_mode = self.device.recording_settings.mode
+            await self.device.set_privacy(True, 0, RecordingMode.NEVER)
```
bdraco, line 178:
```suggestion
            await self.device.set_privacy(True, 0, RecordingMode.NEVER)
            return
```

custom_components/unifiprotect/switch.py
```diff
@@ -118,67 +126,73 @@ class UnifiProtectSwitch(UnifiProtectEntity, SwitchEntity):
...
 
     async def async_turn_off(self, **kwargs):
         """Turn the device off."""
+
+        if isinstance(self.device, Camera):
```
bdraco, line 183: Could save some indent by checking `_KEY_STATUS_LIGHT` first and then reversing this conditional and returning

custom_components/unifiprotect/select.py
```diff
@@ -237,22 +286,36 @@ async def async_select_option(self, option: str) -> None:
...
+                    await self.device.set_lcd_text(None)
+                else:
+                    await self.device.set_lcd_text(DoorbellMessageType(unifi_value))
+
+                _LOGGER.debug("Changed Doorbell LCD Text to: %s", option)
+                return
+
+        if self.entity_description.ufp_enum_type is not None:
+            unifi_value = self.entity_description.ufp_enum_type(unifi_value)
+        elif _KEY_VIEWER:
```
bdraco, line 316: Missing conditional?

custom_components/unifiprotect/camera.py
AngellusMortis: `device.id`  would probably work better hear since the MAC address is unique to the device, which other integrations could use (unless unique IDs for entities are namespace to the integration?). Also, changing it here would make the camera entities inconsistent with every other entity since they are all `id_mac` and we cannot really change those now (that would be a major breaking change).

bdraco:
Finished the additional run-through.   Some of this was pre-existing that I should have caught earlier. Sorry about that.

custom_components/unifiprotect/camera.py
AngellusMortis: I will investigate. This is a fixed size list, so as long as the IDs are consistent within the list, it should be safe to just hardcode the ID -> index map.

custom_components/unifiprotect/camera.py
bdraco: unique ids are name spaced to the integration and platform

https://github.com/home-assistant/core/blob/dev/homeassistant/components/homekit/aidmanager.py#L35

bdraco: We can always migrate them later. 
This one isn't something we need to worry about now.

AngellusMortis: Yeah, I can add the migrate as a follow up PR. Do you have the docs on how to migrate? If I can do that, I can _remove_ the breaking change for the new cameras.

bdraco: https://github.com/home-assistant/core/blob/dev/homeassistant/components/powerwall/__init__.py#L48


and a more complex one


https://github.com/home-assistant/core/blob/dev/homeassistant/components/tplink/migration.py

custom_components/unifiprotect/camera.py
AngellusMortis: It looks like the ids always match the indicies (0, 1, 2) and the number channels return is always 3 regradless if RTSP is enabled. I will update this to just use the id = index.

AngellusMortis:
Pushed up the changes.

custom_components/unifiprotect/binary_sensor.py
AngellusMortis: I moved all of the code in this property to a separate method that is called by `_async_update_device_from_protect`. Then this just uses those values (merging them with the parent ones).

custom_components/unifiprotect/binary_sensor.py
```diff
@@ -299,57 +359,4 @@ def _async_fire_motion_event(self):
...
-
-        attr.update(
-            {
-                ATTR_LAST_TRIP_TIME: self.device.last_motion.isoformat(),
-                ATTR_EVENT_SCORE: score,
-                ATTR_EVENT_OBJECT: detected_object,
-            }
-        )
-        return attr
+        return {**super().extra_state_attributes, **self._attr_extra_state_attributes}
```
bdraco, line 362: Entities are either supposed to implement `self._attr_extra_state_attributes` OR `extra_state_attributes` but not both
https://github.com/home-assistant/core/blob/dev/homeassistant/helpers/entity.py#L318

custom_components/unifiprotect/camera.py
bdraco: Do you want to do that in this PR?  If so I'll hold off testing since the new entities would be created if I do

custom_components/unifiprotect/binary_sensor.py
AngellusMortis: Technically this is both? Lol. The UFP base entity class overrides calling the HA base class. 

https://github.com/briis/unifiprotect/blob/master/custom_components/unifiprotect/entity.py#L57

custom_components/unifiprotect/camera.py
AngellusMortis: Nah, I will do it in a follow up PR. This one is already pretty large as is. As of right now only the main camera entities will be broken if you install this version on a HA instance that was running 0.10.

AngellusMortis: The migration _should_ be to test. Either the entities are orphaned or not. So, it should in comparison be a super simple PR to test.

bdraco: Ideally the existing ones get migrated before the new ones are created so users won't loose their custom entity ids and names

custom_components/unifiprotect/binary_sensor.py
bdraco: It only supposed to be handled by the base entity class.

custom_components/unifiprotect/camera.py
AngellusMortis: I was planning on doing the follow up PR before we release the first beta. So only the 3 of us should be affected. Also, merging this PR now should have no effect on the migration. Already have an idea for the logic and it should be the same with or without this PR.

custom_components/unifiprotect/binary_sensor.py
AngellusMortis: I renamed it.

bdraco:
I tried setting this up fresh and got a trace
```
2021-11-29 17:41:35 ERROR (MainThread) [aiohttp.server] Error handling request
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/aiohttp/web_protocol.py", line 430, in _handle_request
    resp = await request_handler(request)
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/aiohttp/web_app.py", line 504, in _handle
    resp = await handler(request)
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/aiohttp/web_middlewares.py", line 119, in impl
    return await handler(request)
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/homeassistant/components/http/security_filter.py", line 60, in security_filter_middleware
    return await handler(request)
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/homeassistant/components/http/forwarded.py", line 98, in forwarded_middleware
    return await handler(request)
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/homeassistant/components/http/request_context.py", line 24, in request_context_middleware
    return await handler(request)
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/homeassistant/components/http/ban.py", line 78, in ban_middleware
    return await handler(request)
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/homeassistant/components/http/auth.py", line 138, in auth_middleware
    return await handler(request)
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/homeassistant/components/http/view.py", line 137, in handle
    result = await result
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/homeassistant/components/config/config_entries.py", line 157, in post
    return await super().post(request, flow_id)
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/homeassistant/components/http/data_validator.py", line 62, in wrapper
    result = await method(view, request, *args, **kwargs)
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/homeassistant/helpers/data_entry_flow.py", line 110, in post
    result = await self._flow_mgr.async_configure(flow_id, data)
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/homeassistant/data_entry_flow.py", line 252, in async_configure
    result = await self._async_handle_step(flow, cur_step["step_id"], user_input)
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/homeassistant/data_entry_flow.py", line 325, in _async_handle_step
    result: FlowResult = await getattr(flow, method)(user_input)
  File "/Users/bdraco/home-assistant/config/custom_components/unifiprotect/config_flow.py", line 151, in async_step_user
    nvr_data, errors = await self._async_get_nvr_data(user_input)
  File "/Users/bdraco/home-assistant/config/custom_components/unifiprotect/config_flow.py", line 92, in _async_get_nvr_data
    nvr_data = await protect.get_nvr()
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/pyunifiprotect/api.py", line 753, in get_nvr
    return NVR.from_unifi_dict(**data, api=self)
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/pyunifiprotect/data/base.py", line 94, in from_unifi_dict
    data = cls.unifi_dict_to_dict(data)
  File "/Users/bdraco/home-assistant/venv/lib/python3.9/site-packages/pyunifiprotect/data/nvr.py", line 497, in unifi_dict_to_dict
    data["recordingRetentionDuration"] = timedelta(milliseconds=data.pop("recordingRetentionDurationMs"))
TypeError: unsupported type for timedelta milliseconds component: NoneType
```

AngellusMortis:
Is that a device without a HDD? It would make sense if so. Not tested pyunifprotect on devices with no recording retention.

bdraco:
> Is that a device without a HDD? It would make sense if so. Not tested pyunifprotect on devices with no recording retention.

Its a UNVR

AngellusMortis:
PR is up for the fix: https://github.com/briis/pyunifiprotect/pull/112

The PR should definitely fix the recording retention issue, but just in case there is another issue, probably want to test directly against master after the merge before cutting a new pyunifiprotect release. Just to cut down on the release spam.

bdraco:
I'm still testing this. Its taken a while to rename entities to get everything in a testable state

AngellusMortis:
Did you have to rename anything other than the main camera entities? Those are the only ones I expected to break.

bdraco:
So far everything has been working.  I have noticed cpu usage is roughly double what is was before

bdraco:
> Did you have to rename anything other than the main camera entities? Those are the only ones I expected to break.

Nope.  There are 42 of them total between instances so it takes a while

bdraco:
Screen Shot 2021-12-01 at 15 48 06

bdraco:
Should it be fetching bootstrap if the websocket is connected?

bdraco:
Protect went from roughly 8% of the event loop time to ~47% so almost a 6x increase in processing time

AngellusMortis:
Technically yes. We can remove that though if we think it is stable enough. 

It still does a `get_boostrap` on the DEVICE_UPDATE_INTERNAL`. We can either make this a _much_ larger internal (maybe once every 5/10 minutes?) or just get rid of it.

https://github.com/briis/pyunifiprotect/blob/master/pyunifiprotect/api.py#L48
https://github.com/briis/pyunifiprotect/blob/master/pyunifiprotect/api.py#L522

AngellusMortis:
> Protect went from roughly 8% of the event loop time to ~47% so almost a 6x increase in processing time

It looks like from that screenshot that is from the WS updates, right? Which that looks like it is mostly update the bootstrap. I know we are processing _a lot_ more messages now because the whole bootstrap is updated in real time. So, we could stop updating parts of the bootstrap (like the NVR, that updates _constantly_ and we do not use any of the data except the doorbell message stuff).

bdraco:
> > Protect went from roughly 8% of the event loop time to ~47% so almost a 6x increase in processing time
> 
> It looks like from that screenshot that is from the WS updates, right? Which that looks like it is mostly update the bootstrap. I know we are processing _a lot_ more messages now because the whole bootstrap is updated in real time. So, we could stop updating parts of the bootstrap (like the NVR, that updates _constantly_ and we do not use any of the data except the doorbell message stuff).

Ideally we could pre-filter out the data out that we don't need so we can avoid processing it.

bdraco:
This is the most expensive branch

Screen Shot 2021-12-01 at 16 10 24

AngellusMortis:
So yeah, `update_from_dict` can be improved to filter out a lot of the "spam" messages. uptimes, storage stats, etc. Those come in a very large rate.

That would likely be a follow up PR though since it would require mostly changes to `pyunifiprotect`.

bdraco:
How often is this data actually changing?   I wonder if we can also avoid constructing if its the same is the last one

```
[<class 'pyunifiprotect.data.devices.ISPSettings'>,
 None,
 {'ae_mode': 'auto',
  'brightness': 50,
  'contrast': 50,
  'd_zoom_center_x': 50,
  'd_zoom_center_y': 50,
  'd_zoom_scale': 0,
  'd_zoom_stream_id': 4,
  'denoise': 50,
  'focus_mode': 'ztrig',
  'focus_position': 0,
  'hue': 50,
  'icr_sensitivity': 0,
  'ir_led_level': 255,
  'ir_led_mode': <IRLEDMode.AUTO: 'auto'>,
  'is_3dnr_enabled': True,
  'is_aggressive_anti_flicker_enabled': False,
  'is_auto_rotate_enabled': False,
  'is_external_ir_enabled': False,
  'is_flipped_horizontal': False,
  'is_flipped_vertical': False,
  'is_ldc_enabled': True,
  'is_pause_motion_enabled': False,
  'saturation': 50,
  'sharpness': 50,
  'touch_focus_x': 1001,
  'touch_focus_y': 1001,
  'wdr': 1,
  'zoom_position': 0}]
```

AngellusMortis:
I think the metrics (uptime, storage stats, etc. come in every second. Just about. There is an NVR update message about every second for sure.

And yeah, I could definitely optimize `update_from_dict` as well to not reconstruct the _whole_ object every time if only a child UFP object was updated.

AngellusMortis:
And I am already caching the current pre-updated state as well, but that should not be needed since the WS message specifically only gives you the changed data. So, I could iterate the changed keys and update the attributes one at a time instead of using a construct.

AngellusMortis:
> > Should it be fetching bootstrap if the websocket is connected?
> 
> Technically yes. We can remove that though if we think it is stable enough.
> 
> It still does a `get_boostrap` on the `DEVICE_UPDATE_INTERNAL`. We can either make this a _much_ larger internal (maybe once every 5/10 minutes?) or just get rid of it.

@bdraco, thoughts on this one? I was about to go work on improving pyunifiprotect to optimize and improve these calls. What to see what you thought about that one. I think changing it from 60 seconds to 5 minutes is probably a great call with the Websocket running. Then even if we add more sensors to track things that do not get updated in real time, they will get updated at least once every 5 minutes.

bdraco:
> > > Should it be fetching bootstrap if the websocket is connected?
> 
> > 
> 
> > Technically yes. We can remove that though if we think it is stable enough.
> 
> > 
> 
> > It still does a `get_boostrap` on the `DEVICE_UPDATE_INTERNAL`. We can either make this a _much_ larger internal (maybe once every 5/10 minutes?) or just get rid of it.
> 
> 
> 
> @bdraco, thoughts on this one? I was about to go work on improving pyunifiprotect to optimize and improve these calls. What to see what you thought about that one. I think changing it from 60 seconds to 5 minutes is probably a great call with the Websocket running. Then even if we add more sensors to track things that do not get updated in real time, they will get updated at least once every 5 minutes.

Since it's easy to adjust later, I would start with 15m and we can adjust if we get feedback that there is something we need it update it faster.

AngellusMortis:
@bdraco I added the profile WS service I was talking about. Can you do another test run to see how the CPU is looking now? If it is still high, can you do a profile and optionally give the output JSON file?

Hopefully everything will look good enough and then we can release another pyunifiprotect version.

image

bdraco:
```
------------------------------- ws stat summary -------------------------------
packet count: 179
filtered packet count: 4 (97.77%)
--------------------------------------------------------------------------------
camera: 3
event: 1
--------------------------------------------------------------------------------
update: 3
add: 1
--------------------------------------------------------------------------------
last_motion: 3
model: 1
id: 1
type: 1
start: 1
end: 1
score: 1
heatmap_id: 1
camera_id: 1
smart_detect_types: 1
--------------------------------------------------------------------------------
```

bdraco:
I think this is all good now, but I want to test it on my instance that doesn't have asyncio debug on

Screen Shot 2021-12-03 at 08 23 46

Most expensive thing now seems to be the `deepcopy`.

AngellusMortis:
The deepcopy is also just from copying the packet data because they were being modified after the fact. I could probably even clean that up to not copy it if it is not capturing the data.

bdraco (approved):
Testing looks great.  Nice work 👍

bdraco:
Needs a new `pyunifiprotect` release before merge.

bdraco (approved):
Testing looks good

Issue #358, opened 2021-11-27
Exposed wrong settings bricked g3 flex cameras.


incarvr6:
### UniFi Protect HA 0.10.0

After using on g3 flex new exposed setting for high FPS 2 cameras freeze. It is still connected to network and access to local IP is available but no settings are available. After several resets ( including phisical) I managed to get back to life only one.

incarvr6:
![IMG_20211127_230301](https://user-images.githubusercontent.com/49254419/143722779-f2c9b603-56e4-4a62-9cbb-525de807f71b.jpg)
![Screenshot_2021-11-27-21-54-54-27_40deb401b9ffe8e1df2f1cc5ba480b12](https://user-images.githubusercontent.com/49254419/143722790-cba1c73f-fb93-482a-bf06-3836511646e3.jpg)

AngellusMortis:
Can you please update your versions of Home Assistant, UniFi Protect and the HA integration? Latest is not very helpful for troubleshooting.

AngellusMortis:
I do not have G3 Flex cameras, so I might not be of the best help. But I assume you mean the "High FPS mode" switch inside of Home Assistant, right? This _should_ trigger the same update that UFP itself does, so it _should_ be no different than how doing it from the UFP app / WebUI is. I know with my G4 Pros, switching on/off High FPS mode that kind of "lock up" the camera for a few seconds, while it changes everything internally. But it comes back once it is done.

Something to assistant with debugging though, can you do the following and try to get me the outputted data?

1. Go to the UniFi Protect WebUI
2. Click on one of your G3 Flex cameras
3. Go to the Recording Tab. Do you have a "High Frame Rate Mode" under Recording Quality:
![image](https://user-images.githubusercontent.com/490848/143722887-91345001-a469-4670-8a1c-1ff4335dec5f.png)
4. If so turn _off_ High Frame Rate Mode
5. Copy the Camera ID out of the URL (it should look like `<protect_url>/protect/devices/####` where `####` is your Camera ID
6. Go to `<protect_url>/proxy/protect/api/cameras/####`. Copy everything on this page and paste it back in to this issue.
7. If you _do_ have an High Frame Rate Mode (question from #3), go back and turn it back on.
8. Reload the URL from #6 and post the _new_ data here.

incarvr6:
I'm on phone right now so very hard to type and switch between multiple windows.. that's regarding versions.

No on G3 flex or G3 bullet there is no settings for high FPS

[...]

PR #359, opened 2021-11-28
Disables High FPS switch

AngellusMortis:
Temporarily disables High FPS mode switch. We are not doing the proper checks on 0.10 to verify the camera _has_ a High FPS mode before adding the switch. Cameras that do not have one (G3 Flex, G3 Instance, G3 Bullet and G4 Instant) will become bricked and it is quite difficult to fix the issue.

Temporarily fix for https://github.com/briis/unifiprotect/issues/358 until we get 0.11 done.

The PR I have for 0.11 has this issue fixed already as I am using `feature_flags.video_modes` to determine if the camera has a High FPS mode.

Image of High FPS mode switch existing on my G4 Instant: 
image

But as you can see, it does not have a high FPS video mode. 
image

### DIFF
--- a/custom_components/unifiprotect/switch.py
+++ b/custom_components/unifiprotect/switch.py
@@ -58,14 +58,14 @@ SWITCH_TYPES: tuple[UnifiProtectSwitchEntityDescription, ...] = (
         ufp_required_field="has_hdr",
         ufp_value="hdr_mode",
     ),
-    UnifiProtectSwitchEntityDescription(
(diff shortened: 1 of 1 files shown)


AngellusMortis:
@briis When you get the chance, please review this, merge it and cut a new 0.10.x release for it.

It is being merged into the `release/0.10.x` branch, so make sure you check out that branch before making a tag for HACS.

bdraco (approved):

Issue #304, opened 2021-09-29
play custom sound on camera speaker


aaronlower:
feature request to play a custom sound on the camera speaker (e.g. G3 instant, G4 dome, G4 doorbell) when person is detected or another event is triggered

AngellusMortis:
I can confirm I have figured out the `talkbackSettings` and I am implementing an interface for this in `pyunifiprotect` so we can hopefully implement a `media_player` entity for cameras that have speakers / talkback in the "near ish" future (maybe for 0.11?).

AngellusMortis:
Added to the 0.11 beta. Doorbells may still have some issues that I am looking into more, but those may be the result of general doorbell issues everyone else has had post 1.20 as well.

disabledtech:
Hi,

I'm having some issues with the media_player for my G4 doorbell. I can get it to play TTS messages and custom sounds (e.g. a buzzer .mp3) but only a few times before it stops working. To get it playing again I need to reboot the doorbell. After a reboot I can do a few more sound/TTS before it stops working again.

Home Assistant version: 2021.12.3 running on a raspberry pi 4B 8GB
UniFi Protect Integration: 0.11.4 via HACS
UniFi Protect: 1.20.3 running on a UDM Pro
G4 Firmware: 4.46.18

I enable debugging in my config. I see this line every 2 seconds.

`2021-12-29 17:28:36 DEBUG (MainThread) [pyunifiprotect.api] Skipping update since websocket is active`

Here's a longer snippet of the logs directly after trying to use TTS directly through the G4 Doorbell Speaker entity.

```
2021-12-29 17:32:30 DEBUG (MainThread) [pyunifiprotect.api] Skipping update since websocket is active
2021-12-29 17:32:32 DEBUG (MainThread) [pyunifiprotect.api] Skipping update since websocket is active
2021-12-29 17:32:33 DEBUG (MainThread) [custom_components.unifiprotect.media_player] Playing Media http://192.168.1.119:8123/api/tts_proxy/a94a8fe5ccb19ba61c4c0873d391e987982fbbd3_en-us_826d21f2a0_cloud.mp3 for UVC G4 Doorbell Speaker
[...]

PR #361, opened 2021-11-28
Adds Media Player Entities. Overhauls Doorbell Text Select Entities.

AngellusMortis:
* Removes the breaking change for the unique IDs for the cameras changing from previous PR.

* `CHANGE`: **BREAKING CHANGE** The Doorbell Text select entity for Doorbells has been overhauled. The Config Flow option for Doorbell Messages has been removed. You now can use the the  `unifiprotect.add_doorbell_text` and `unifiprotect.remove_doorbell_text` services to add/remove Doorbell messages. This will persist the messages in UniFi Protect and the choices will now be the same ones that appear in the UniFi Protect iOS/Android app.

image
image
image

* `NEW`: Added `unifiprotect.set_default_doorbell_text` service to allow you to set your default Doorbell message text.

* `NEW`: Added "SSH Enabled" switch for all adoptable UniFi Protect devices. This switch is disabled by default.

* `NEW`: Added media player entity for cameras with speaker. Speaker will accept any ffmpeg playable audio file URI (URI must be accessible from _Home Assistant_, not your Camera). TTS works great!
    * Implements #304 

image

* `NEW`: Adds config option to update all metrics (storage stat usage, uptimes, CPU usage, etc.) in realtime. **WARNING**: Enabling this option will greatly increase your CPU usage (~2x is what we were seeing in our testing. It is recommended to leave it disabled for now as we do not have a lot of diagnostic sensors using this data yet.

* `CHANGE`: The Bluetooth strength sensor for the UP Sense is now disabled by default (will not effect anyone that already has the sensor).


custom_components/unifiprotect/button.py
```diff
@@ -0,0 +1,62 @@
...
+        protect: ProtectApiClient,
+        protect_data: UnifiProtectData,
+        device: ProtectAdoptableDeviceModel,
+    ):
+        """Initialize an Unifi camera."""
+
+        super().__init__(protect, protect_data, device, None)
+
+        self._attr_name = f"{self.device.name} Reboot Device"
+        self._attr_entity_registry_enabled_default = False
```
iMicknl, line 55: Not sure if you are aware, but they added device classes for buttons. (example: https://github.com/home-assistant/core/pull/60625/files). Probably not released yet, however maybe good to take into account for a follow up PR / this PR.

```device_class=ButtonDeviceClass.RESTART```

Comment (AngellusMortis) [reply custom_components/unifiprotect/button.py:55]: I did not know that, thanks for the heads up. I will either add it for this PR or make a note of it for the core merge so we do not need to depend on 2021.12 after I dig into it more.

Comment (AngellusMortis) [reply custom_components/unifiprotect/button.py:55]: It looks button entities to begin with are 2021.12+. So, I added the device class and added a check for 2021.12 before initializing the platform. Definitely thanks for the heads up about the device class though.

File: custom_components/unifiprotect/config_flow.py [hunk 2]
```diff
@@ -197,6 +199,10 @@ async def async_step_init(self, user_input=None):
                         CONF_DISABLE_RTSP,
                         default=self.config_entry.options.get(CONF_DISABLE_RTSP, False),
                     ): bool,
+                    vol.Optional(
+                        CONF_ALL_UPDATES,
+                        default=self.config_entry.options.get(CONF_ALL_UPDATES, False),
+                    ): bool,
```
bdraco, line 205: Is there a use case for users turning this one who are not devs?

AngellusMortis: Not yet. But I planned on adding more diagnostic sensors for things like WiFi strength, etc and those would likely require that enabled.

AngellusMortis: Actually, one sensor, _might_ need it now already. I the UP Sense already has a Bluetooth strength sensor.

custom_components/unifiprotect/__init__.py
```diff
@@ -48,6 +58,93 @@
...
+    # migrate entities
+    registry = er.async_get(hass)
+    mac_to_id: dict[str, str] = {}
+    mac_to_channel_id: dict[str, str] = {}
+    bootstrap = await protect.get_bootstrap()
+    for model in DEVICES_WITH_ENTITIES:
+        attr = model.value + "s"
+        for device in getattr(bootstrap, attr).values():
+            mac_to_id[device.mac] = device.id
+            if model == ModelType.CAMERA:
```
bdraco, line 87:
```suggestion
            if model != ModelType.CAMERA:
                continue
```

... and outdent below

custom_components/unifiprotect/__init__.py
```diff
@@ -48,6 +58,93 @@
...
+                device_id = mac_to_id[mac]
+                channel_id = mac_to_channel_id[mac]
+                new_unique_id = f"{device_id}_{channel_id}"
+            else:
+                device_id = parts[0]
+                channel_id = parts[2]
+                extra = "" if len(parts) == 3 else "_insecure"
+                new_unique_id = f"{device_id}_{channel_id}{extra}"
+
+        if new_unique_id is not None:
```
bdraco, line 122:
```suggestion
        if new_unique_id is None:
           continue
```

Then outdent below

custom_components/unifiprotect/data.py
```diff
@@ -90,8 +91,22 @@ async def async_refresh(self, *_, force=False):
...
+                self.async_signal_device_id_update(viewer.id)
+        # trigger update for all Cameras with LCD screens when NVR Doorbell settings updates
+        elif (
+            isinstance(message.new_obj, NVR)
+            and "doorbell_settings" in message.changed_data
+        ):
+            self._protect.bootstrap.nvr.update_all_messages()
+            for camera in self._protect.bootstrap.cameras.values():
+                if camera.feature_flags.has_lcd_screen:
+                    self.async_signal_device_id_update(camera.id)
```
bdraco, line 109: This is starting to get a bit long, can we avoid an `isinstance ` linear search?

custom_components/unifiprotect/select.py
```diff
@@ -242,16 +202,47 @@ def __init__(
...
+            self._attr_options = [item["name"] for item in options]
+            self._hass_to_unifi_options = {item["name"]: item["id"] for item in options}
+            self._unifi_to_hass_options = {item["id"]: item["name"] for item in options}
+        self._async_update_device_from_protect()
+
+    @callback
+    def _async_update_device_from_protect(self):
+        super()._async_update_device_from_protect()
+
+        if self.entity_description.key in (_KEY_VIEWER, _KEY_DOORBELL_TEXT):
```
bdraco, line 223:
```suggestion
        if self.entity_description.key not in (_KEY_VIEWER, _KEY_DOORBELL_TEXT):
           return
```
outdent below

custom_components/unifiprotect/select.py
```diff
@@ -242,16 +202,47 @@ def __init__(
...
+                )
+
+                options = [
+                    {"id": "", "name": f"Default Message ({default_message})"},
+                    *built_messages,
+                ]
+
+            self._attr_options = [item["name"] for item in options]
+            self._hass_to_unifi_options = {item["name"]: item["id"] for item in options}
+            self._unifi_to_hass_options = {item["id"]: item["name"] for item in options}
```
bdraco, line 245: Options aren't allowed to change after the entity has been initialized since integrations that consume them once (ie Alex, HomeKit, Google Assistant, etc) will break

AngellusMortis: Is there a better way to handle this then?  Liveviews and the Doorbell texts are by their nature very dynamic, and it is rather annoying to have to reload UFP/restart HA to get the new options.

custom_components/unifiprotect/__init__.py
```diff
@@ -110,17 +208,46 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
...
+        message: str = call.data[CONF_MESSAGE]
+        await protect.bootstrap.nvr.add_custom_doorbell_message(message)
+
+    async def remove_doorbell_text(call: ServiceCall):
+        message: str = call.data[CONF_MESSAGE]
+        await protect.bootstrap.nvr.remove_custom_doorbell_message(message)
+
+    async def set_default_doorbell_text(call: ServiceCall):
+        message: str = call.data[CONF_MESSAGE]
+        await protect.bootstrap.nvr.set_default_doorbell_message(message)
```
bdraco, line 230: These should be entity services since they can have multiple protect installs

https://developers.home-assistant.io/docs/dev_101_services/#entity-services

custom_components/unifiprotect/data.py
AngellusMortis: I cannot think of a good way. Each one is just different enough. The list is not likely to get any larger though.

custom_components/unifiprotect/__init__.py
```diff
@@ -110,17 +208,46 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
...
 
     async def profile_ws(call: ServiceCall):
```
bdraco, line 232: This one should probably as well (target the device -- examples in homekit services.yaml)

custom_components/unifiprotect/select.py
bdraco: I can't think of a better way right now.  I agree it is annoying, but it is what we have right now.  It needs to be improved int he base architecture

AngellusMortis: Persistent notification to alert the user they have changed, and they need to reload the integration maybe?

bdraco: Persistent notifications should only be shown for high importance items that the user is currently interacting with or need to take action soon so that seems a bit too much.

For other integrations we have logged an error that the sources/data is out of sync and a restart is required to get everything back in sync.

bdraco (approved):

PR #366, opened 2021-11-30
Updates docs in various places for 2021.11

AngellusMortis:


briis:
Looks good to me. Don't spend a lot of time on the `unifiprotect.markdown` document. This was just a start of making the official documentation for the Core Integration. So far from anything that is useful right now.

AngellusMortis:
Yep, I figured. That is why I removed all of the `unifi` stuff since we will be updating it a lot more in the future.

PR #376, opened 2021-12-04
Improves service calls

AngellusMortis:
* Allows `device_id` for service calls to be any device from the UFP instance.
* Fixes issues with number entity from previous merge

### DIFF
--- a/custom_components/unifiprotect/services.py
+++ b/custom_components/unifiprotect/services.py
@@ -6,7 +6,9 @@ from homeassistant.core import HomeAssistant, ServiceCall
 from homeassistant.exceptions import HomeAssistantError
 from homeassistant.helpers import device_registry as dr
 from homeassistant.helpers.service import async_extract_referenced_entity_ids
+from pydantic import ValidationError
 from pyunifiprotect.api import ProtectApiClient
+from pyunifiprotect.exceptions import BadRequest
 
 from .const import CONF_DURATION, CONF_MESSAGE, DOMAIN
 from .models import UnifiProtectEntryData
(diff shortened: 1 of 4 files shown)


custom_components/unifiprotect/services.py
```diff
@@ -22,65 +24,101 @@ def _async_all_ufp_instances(hass: HomeAssistant) -> list[ProtectApiClient]:
     ]
 
 
+def _async_get_macs_for_device(device_entry: dr.DeviceEntry) -> list[str]:
+    return [
+        # MAC addresses in UFP are always caps
+        cval.replace(":", "").upper()
```
bdraco, line 30: use dr.format_mac

custom_components/unifiprotect/services.py
```diff
@@ -22,65 +24,101 @@ def _async_all_ufp_instances(hass: HomeAssistant) -> list[ProtectApiClient]:
...
+
+    try:
+        await asyncio.gather(
+            *(
+                i.bootstrap.nvr.set_default_doorbell_message(message)
+                for _, i in instances
+            )
+        )
+    except (BadRequest, ValidationError) as err:
+        raise HomeAssistantError(str(err)) from err
```
bdraco, line 113: Similar code to above.  Can we make it its own function and pass in the coro to call?

custom_components/unifiprotect/services.py
AngellusMortis: This is actually to revert the formatting of the mac so it matches the UFP value. The addresses for UFP are always capital with no `:`. 

image

bdraco: Or do it the other way around and run `i.bootstrap.nvr.mac` though `dr.format_mac`

custom_components/unifiprotect/services.py
```diff
@@ -22,65 +24,92 @@ def _async_all_ufp_instances(hass: HomeAssistant) -> list[ProtectApiClient]:
     ]
 
 
+def _async_unifi_mac_from_hass(mac: str):
```
bdraco, line 27:
```suggestion
@callback
def _async_unifi_mac_from_hass(mac: str) -> str:
```

custom_components/unifiprotect/services.py
```diff
@@ -22,65 +24,92 @@ def _async_all_ufp_instances(hass: HomeAssistant) -> list[ProtectApiClient]:
...
+    # MAC addresses in UFP are always caps
+    return mac.replace(":", "").upper()
+
+
+def _async_get_macs_for_device(device_entry: dr.DeviceEntry) -> list[str]:
```
bdraco, line 32:
```suggestion
@callback
def _async_get_macs_for_device(device_entry: dr.DeviceEntry) -> list[str]:
```

I don't think we ever pass these to `hass.add_job` but if we did they would run in the executor without the decortator

bdraco (approved):

bdraco (approved):

bdraco (approved):

PR #379, opened 2021-12-05
Adds NVR sensors

AngellusMortis:
* `NEW`: Adds NVR sensors. All of them are disabled by default. All of the sensors will only update every ~15 minutes unless the "Realtime metrics" config option is turned on. List of all sensors:
    * Disk Health (one per disk)
    * System Info: CPU Temp, CPU, Memory and Storage Utilization
    * Uptime
    * Recording Capacity (in seconds)
    * Distributions of stored video for Resolution (4K/HD/Free)
    * Distributions of stored video for Type (Continuous/Detections/Timelapse)

What the NVR device actually looks like now with all of them turned on:
image


custom_components/unifiprotect/binary_sensor.py
```diff
@@ -146,6 +164,38 @@ async def async_setup_entry(
 ) -> None:
     """Set up binary sensors for UniFi Protect integration."""
     entry_data: UnifiProtectEntryData = hass.data[DOMAIN][entry.entry_id]
+    entities = _async_device_entities(entry_data)
+    entities = _async_nvr_entities(entry_data)
+
+    async_add_entities(entities)
+
```
bdraco, line 171: Variable defined twice

bdraco:
Change the uptime sensor to a boot time sensor so it doesn't constant generate state change events.

AngellusMortis:
> Change the uptime sensor to a boot time sensor so it doesn't constant generate state change events.

It actually already is. There is `uptime` and `up_since` from UFP. I am using `up_since`, but still calling them "uptime" sensors.

custom_components/unifiprotect/binary_sensor.py
AngellusMortis: Haha. Oops. I guess I did not test the other binary sensors. They were probably all broken. I should make sure I do that.

AngellusMortis:
It does look like it is constantly changing though. I think I saw this somewhere else. I need to truncate the milliseconds because they seem to always constantly change.
image

custom_components/unifiprotect/sensor.py
```diff
@@ -157,25 +298,62 @@ def __init__(
...
+
+        if isinstance(new_value, datetime):
+            # protection to prevent uptime from changing if UniFi Protect changes values slightly
+            if hasattr(self, "_attr_native_value") and self._attr_native_value:
+                diff = abs((self._attr_native_value - new_value).total_seconds())
+                if diff < 5:
+                    return
+
+            if not above_ha_version(2021, 12):
+                new_value = new_value.isoformat()
```
AngellusMortis, line 351: Thoughts on this? I changed it to use `_attr_native_value` so I can see the previous value and then require the diff to be more then 5 seconds before updating. My G4 on master kept getting new values even though I had `replace(microsecond=0)` so that must mean the second value was slightly changing over time as well. It was consistently Decmember 3rd at 10:26 PM though.

custom_components/unifiprotect/binary_sensor.py
```diff
@@ -146,6 +164,38 @@ async def async_setup_entry(
 ) -> None:
     """Set up binary sensors for UniFi Protect integration."""
     entry_data: UnifiProtectEntryData = hass.data[DOMAIN][entry.entry_id]
+    entities = _async_device_entities(entry_data)
+    entities += _async_nvr_entities(entry_data)
+
+    async_add_entities(entities)
+
+
+def _async_nvr_entities(
```
bdraco, line 173:
```suggestion
@callback
def _async_nvr_entities(
```

custom_components/unifiprotect/binary_sensor.py
```diff
@@ -146,6 +164,38 @@ async def async_setup_entry(
...
+            )
+            _LOGGER.debug(
+                "Adding binary sensor entity %s",
+                (description.name or "{index}").format(index=index),
+            )
+
+    return entities
+
+
+def _async_device_entities(
```
bdraco, line 196:
```suggestion
@callback
def _async_device_entities(
```

Not strictly needed

custom_components/unifiprotect/binary_sensor.py
```diff
@@ -268,6 +325,20 @@ def _async_get_extra_attrs(self) -> dict[str, Any]:
...
+            assert isinstance(self.device, NVR)
+            assert self._index is not None
+
+            disks = self.device.system_info.storage.devices
+            self._attr_available = self._attr_available and len(disks) > self._index
+            if self._attr_available:
+                disk = disks[self._index]
+                self._attr_is_on = not disk.healthy
+
+                self._extra_state_attributes = {ATTR_MODEL: disk.model}
```
bdraco, line 338: Don't not need to check `_attr_available`, the state will be calculated automatically
```suggestion
             disk = disks[self._index]
             self._attr_is_on = not disk.healthy
             self._extra_state_attributes = {ATTR_MODEL: disk.model}
```

AngellusMortis: This is on purpose to avoid an index error. I am setting `_attr_avaiable` right above with if the index exists.

custom_components/unifiprotect/binary_sensor.py
```diff
@@ -268,6 +325,20 @@ def _async_get_extra_attrs(self) -> dict[str, Any]:
     def _async_update_device_from_protect(self) -> None:
         super()._async_update_device_from_protect()
 
+        if self.entity_description.key.startswith(_KEY_DISK_HEALTH):
+            assert isinstance(self.device, NVR)
+            assert self._index is not None
+
+            disks = self.device.system_info.storage.devices
+            self._attr_available = self._attr_available and len(disks) > self._index
+            if self._attr_available:
```
bdraco, line 334:
```suggestion
            if self._index in disks:
```

bdraco: I think this shows the intent a bit more clearly

custom_components/unifiprotect/binary_sensor.py
bdraco: https://github.com/briis/unifiprotect/pull/379/files#r762597650

custom_components/unifiprotect/entity.py
```diff
@@ -31,22 +33,35 @@ def __init__(
...
+            assert isinstance(self.device, ProtectAdoptableDeviceModel)
+            self._attr_device_info = DeviceInfo(
+                name=self.device.name,
+                manufacturer=DEFAULT_BRAND,
+                model=self.device.type,
+                via_device=(DOMAIN, self.protect.bootstrap.nvr.mac),
+                sw_version=self.device.firmware_version,
+                connections={(dr.CONNECTION_NETWORK_MAC, self.device.mac)},
+                configuration_url=self.device.protect_url,
+            )
```
bdraco, line 64: Can we avoid an `if` here by using a different base class for `NVR` entities?

custom_components/unifiprotect/sensor.py
```diff
@@ -157,25 +297,62 @@ def __init__(
...
-
-        assert isinstance(value, (str, int, float, datetime)) or value is None
-        if isinstance(value, datetime) and not above_ha_version(2021, 12):
-            return value.replace(microsecond=0).isoformat()
+    @callback
+    def _async_nvr_value(self) -> float:
+        assert isinstance(self.device, NVR)
+        if self.entity_description.key == _KEY_MEMORY:
+            memory = self.device.system_info.memory
+            value = (1 - memory.available / memory.total) * 100
```
bdraco, line 313:
```suggestion
        if self.entity_description.key == _KEY_MEMORY:
            memory = self.device.system_info.memory
            return (1 - memory.available / memory.total) * 100
```
Drop the `else`
Then outdent below.

custom_components/unifiprotect/entity.py
AngellusMortis: That would require bubbling that up to Binary Sensors and Sensors as well though. But I was thinking about possibly look into a way to make clean up making the entities.

custom_components/unifiprotect/sensor.py
```diff
@@ -157,25 +297,62 @@ def __init__(
...
+
+        if isinstance(new_value, timedelta):
+            new_value = new_value.total_seconds()
+
+        if isinstance(new_value, datetime):
+            # protection to prevent uptime from changing if UniFi Protect changes values slightly
+            if hasattr(self, "_attr_native_value") and self._attr_native_value:
+                diff = abs((self._attr_native_value - new_value).total_seconds())
+                if diff < 5:
+                    return
```
bdraco, line 347: Can we prevent the wobbly time from getting into the integration by cleaning this in up in `pyunifiprotect`?

AngellusMortis: Not in a great way. The way keys are currently being filtered only has the incoming value, not the existing. By the time you can compare existing to new, the `changed_data` has already been generated for the subscription message which would cause a diff between `changed_data` and `new_obj`.

custom_components/unifiprotect/sensor.py
AngellusMortis: I actually do not need like any of that anymore. That was the old code before I make the pyunifiprotect changes.

custom_components/unifiprotect/entity.py
AngellusMortis: Yeah, I think this is the best it can be for now. I really like how binary_sensor.py works though and would love to generalize that pattern for all of the entities to make it easier to use different flows/classes for different device types.

custom_components/unifiprotect/sensor.py
AngellusMortis: I could also just truncate the seconds, but then that will lose a bit of resolution into when it came online.

bdraco: That seems like a good compromise for now.

AngellusMortis: Just to clarify, to truncate the seconds?

bdraco: Yes

custom_components/unifiprotect/binary_sensor.py
```diff
@@ -131,6 +138,17 @@ class UnifiProtectBinaryEntityDescription(
     ),
 )
 
+DISK_SENSORS: tuple[UnifiProtectBinaryEntityDescription, ...] = (
+    UnifiProtectBinaryEntityDescription(
+        key=_KEY_DISK_HEALTH,
+        name="Disk {index} Health",
+        device_class=DEVICE_CLASS_PROBLEM,
```
bdraco, line 145: We should set ENTITY_CATEGORY_DIAGNOSTIC

bdraco:
Everything showing Unknown on initial load... will wait 15 to see if it populates
Screen Shot 2021-12-05 at 14 37 34

AngellusMortis:
Opps on the unknown. I forgot to call the method to initialize their values.

bdraco:
Screen Shot 2021-12-05 at 15 05 52

Probably makes sense to drop the float on the recording capacity to seconds, and round everything else to 2 places

bdraco (approved):
LGTM 👍 
Screen Shot 2021-12-05 at 15 54 47

PR #381, opened 2021-12-06
Adds event thumbnail proxy view

AngellusMortis:
* Adds `http` component as a dependency (for the new view).

* Refactors out methods to retrieve a UFP instance from a device so it can be used elsewhere.

* `NEW`: Adds event thumbnail proxy view.
  * URL is `/api/ufp/thumbnail/{thumb_id}`. `thumb_id` is the ID of the thumbnail from UniFi Protect.
  * `device_id` is a required query parameters. `device_id` must be the UniFi Protect device ID for Camera entity
  * `token` is a required query parameter is you are _not_ authenticated. It is an attribute on the motion sensor for the Camera
  * `w` and `h` are optional query string params for thumbnail resizing.

* `NEW`: Adds `event_thumbnail` attribute to Motion binary sensor that uses above mentioned event thumbnail proxy view.

Note about testing: because the attribute is going to change very quickly, the best it with a debugger or automation. You can test the Thumbnail view going to `/proxy/protect/api/events?type=smartDetectZone&limit=100` or `/proxy/protect/api/events?type=motion&limit=100` in your UFP Instance and then pull a `thumbnail` value and use that in as the `thumb_id` param. `entity_id` can be any entity from that UFP instance or you can use `nvr_id` as the ID from `/proxy/protect/api/nvr` in UFP.


custom_components/unifiprotect/views.py
```diff
@@ -0,0 +1,65 @@
...
+
+from .utils import get_ufp_instance_from_entity_id
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class ThumbnailProxyView(HomeAssistantView):
+    """View to proxy event thumbnails from UniFi Protect"""
+
+    requires_auth = False
```
bdraco, line 20: Why no auth?

AngellusMortis: The plan is to use this in Push Notification URLs so they cannot require auth. The `thumb_id` is already random 24-character value. Should be really hard to brute force. 

I can add another randomized value if you think that is not good enough.

AngellusMortis: Actually, looking at the event IDs more closely. They do not seem completely randomize. They seem to be incrementing in some way. I definitely think I need to add another value to the URL to increase entropy. Let me look how camera_proxy works.

AngellusMortis: After looking at how camera_proxy works and thinking about, I think I am just going to reimplement the camera_proxy pattern onto the motion binary sensors.

I was also planning on using this view later for a media browser to view events and clips, but I think I will make a subclass of this view (or vice versa) and have one that does not require auth/has access token and one that does.

AngellusMortis: Okay, I added the access token pattern that `camera_proxy` uses. Same general idea, but now the URL also uses all UFP values outside of the access token.

custom_components/unifiprotect/binary_sensor.py
```diff
@@ -165,12 +174,48 @@ async def async_setup_entry(
...
+                description.name,
+                device.name,
+            )
+
+    @callback
+    def update_tokens(time: datetime) -> None:
+        """Update tokens of the entities."""
+        for entity in entities:
+            entity.async_update_token()
+            entity.async_write_ha_state()
```
bdraco, line 211: We aren't allowed to call `async_write_ha_state` from outside the entity. This needs to use `async_dispatcher_send` / `async_dispatcher_connect`.  (I'm aware the camera integration is using the old method)

AngellusMortis: I am looking how Sonos does it since they seem to have something similar where they poll the speakers for data. It sounds like this might be the flow?

* On set up entry, do a `async_dispatcher_connect` for a custom event with `update_tokens` as the callback.
* using the `async_track_time_interval` method I am here, fire the update tokens event on an interval.

custom_components/unifiprotect/data.py
```diff
@@ -168,3 +172,12 @@ def async_signal_device_id_update(self, device_id: str) -> None:
...
             update_callback()
+
+    @callback
+    def async_add_access_token_entity(self, entity: AccessTokenMixin) -> None:
+        # AccessTokenMixin is always an instance of UnifiProtectEntity
+        self._access_token_entities[entity.device.id] = entity  # type: ignore
+
+    @callback
+    def async_get_access_tokens_entity(self, entity_id: str) -> AccessTokenMixin | None:
+        return self._access_token_entities.get(entity_id)
```
bdraco, line 183: Storing entities in a shared container is not allowed.

AngellusMortis: That is what `homeassistant.helpers.entity_component.EntityComponent`. Since we have our own data storage class, I was just kind of reimplementing that. Otherwise, I need to get the actually real entity class (not the `ReigstryEntity`) to get the access tokens.

bdraco: I think you can work around this by having an object to class the access tokens so the class objects end up in a shared container but the entities themselves don't

AngellusMortis: Maybe should I instead store the access tokens on `protect_data` and then fetch them into the entity? Since it had a reference to protect data.

bdraco: I think that solves it as well

custom_components/unifiprotect/binary_sensor.py
```diff
@@ -441,3 +441,70 @@ def _async_fire_motion_event(self) -> None:
...
+
+        thumb_url: str | None = None
+        if event is not None:
+            assert self.device_info is not None
+            # thumbnail_id is never updated via WS, but it is always e-{event.id}
+            thumb_url = (
+                ThumbnailProxyView.url.format(event_id=f"e-{event.id}")
+                + f"?entity_id={self.entity_id}"
+            )
+            _LOGGER.debug("Event thumbnail URL for %s: %s", self.entity_id, thumb_url)
```
bdraco, line 498: Side effects aren't allowed in properties

custom_components/unifiprotect/binary_sensor.py
```diff
@@ -441,3 +441,70 @@ def _async_fire_motion_event(self) -> None:
...
+            score = self.device.last_smart_detect_event.score
+            event = self.device.last_smart_detect_event
+            detected_object: str | None = (
+                self.device.last_smart_detect_event.smart_detect_types[0].value
+            )
+            _LOGGER.debug(
+                "OBJECTS: %s on %s",
+                self.device.last_smart_detect_event.smart_detect_types,
+                self._attr_name,
+            )
```
bdraco, line 480: Side effects aren't allowed in properties

custom_components/unifiprotect/binary_sensor.py
AngellusMortis: That is not a property (despite what the name my indicate). It is called by `_async_update_device_from_protect`, which is ran as part of the event update for the sensor.

custom_components/unifiprotect/binary_sensor.py
```diff
@@ -441,3 +441,70 @@ def _async_fire_motion_event(self) -> None:
...
+        thumb_url: str | None = None
+        if event is not None:
+            assert self.device_info is not None
+            # thumbnail_id is never updated via WS, but it is always e-{event.id}
+            thumb_url = (
+                ThumbnailProxyView.url.format(event_id=f"e-{event.id}")
+                + f"?entity_id={self.entity_id}"
+            )
+            _LOGGER.debug("Event thumbnail URL for %s: %s", self.entity_id, thumb_url)
+            thumb_url += f"&token={self.access_tokens[-1]}"
```
bdraco, line 499: It would be cleaner to use `from urllib.parse import urlencode`

` urlencode({'entity_id':'x','token':'xxx'})`

custom_components/unifiprotect/data.py
```diff
@@ -168,3 +173,9 @@ def async_signal_device_id_update(self, device_id: str) -> None:
 
         for update_callback in self._subscriptions[device_id]:
             update_callback()
+
+    @callback
+    def async_get_or_create_access_tokens(self, entity_id: str) -> collections.deque:
```
bdraco, line 178: Missing docstring

custom_components/unifiprotect/entity.py
```diff
@@ -99,8 +120,64 @@ def _async_updated_event(self) -> None:
...
         self.async_on_remove(
             self.protect_data.async_subscribe_device_id(
                 self.device.id, self._async_updated_event
             )
         )
+
+
+class AccessTokenMixin(Entity):
+    @property
+    def access_tokens(self) -> collections.deque[str]:
```
bdraco, line 133: Missing docstring

custom_components/unifiprotect/entity.py
```diff
@@ -99,8 +120,64 @@ def _async_updated_event(self) -> None:
...
+
+    @callback
+    def _trigger_update_tokens(self, *args, **kwargs):
+        assert isinstance(self, UnifiProtectEntity)
+        async_dispatcher_send(
+            self.hass,
+            f"{EVENT_UPDATE_TOKENS}-{self.entity_description.key}-{self.entity_id}",
+        )
+
+    async def async_added_to_hass(self) -> None:
```
bdraco, line 167: Missing docstring

custom_components/unifiprotect/views.py
```diff
@@ -0,0 +1,86 @@
...
+        self.hass = hass
+        self.data = hass.data[DOMAIN]
+
+    def _get_access_tokens(
+        self, entity_id: str
+    ) -> tuple[collections.deque, ProtectApiClient] | None:
+
+        for entry in self.data.values():
+            if entity_id in entry.protect_data.access_tokens:
+                return entry.protect_data.access_tokens[entity_id], entry.protect
```
bdraco, line 36: Linear search, but probably doesn't matter since they are unlikely to have many of them

custom_components/unifiprotect/views.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class ThumbnailProxyView(HomeAssistantView):
+    """View to proxy event thumbnails from UniFi Protect"""
+
+    requires_auth = False
+    url = "/api/ufp/thumbnail/{event_id}"
+    name = "api:ufp_thumbnail"
+
+    def __init__(self, hass: HomeAssistant):
```
bdraco, line 26:
```suggestion
    def __init__(self, hass: HomeAssistant) -> None:
```

Missnig docstring

custom_components/unifiprotect/views.py
AngellusMortis: Yeah, that is just if there are multiple Protect Entries. Likely not going to be more than a 1 or 2.

custom_components/unifiprotect/binary_sensor.py
bdraco: This is fine, but consider renaming to avoid future confusion during core review.

custom_components/unifiprotect/binary_sensor.py
```diff
@@ -441,3 +442,65 @@ def _async_fire_motion_event(self) -> None:
...
+        attrs.update(
+            {
+                ATTR_LAST_TRIP_TIME: self.device.last_motion,
+                ATTR_EVENT_SCORE: score,
+                ATTR_EVENT_OBJECT: detected_object,
+                ATTR_EVENT_THUMB: thumb_url,
+            }
+        )
+
+        return attrs
```
bdraco, line 506:
```suggestion
        return (
            {
                **super()._async_update_extra_attrs_from_protect(),
                ATTR_LAST_TRIP_TIME: self.device.last_motion,
                ATTR_EVENT_SCORE: score,
                ATTR_EVENT_OBJECT: detected_object,
                ATTR_EVENT_THUMB: thumb_url,
            }
        )
```

custom_components/unifiprotect/binary_sensor.py
```diff
@@ -441,3 +442,65 @@ def _async_fire_motion_event(self) -> None:
...
+        description: UnifiProtectBinaryEntityDescription,
+        index: int | None = None,
+    ) -> None:
+        assert isinstance(device, Camera)
+        self.device: Camera = device
+        super().__init__(protect, protect_data, device, description, index)
+
+    @callback
+    def _async_update_extra_attrs_from_protect(self) -> dict[str, Any]:
+        attrs = super()._async_update_extra_attrs_from_protect()
```
bdraco, line 462:
```suggestion
```

custom_components/unifiprotect/binary_sensor.py
bdraco: or depending on your preference write as `**attrs`

custom_components/unifiprotect/data.py
```diff
@@ -168,3 +173,10 @@ def async_signal_device_id_update(self, device_id: str) -> None:
...
+        """Wrapper around access_tokens to automatically create underlaying data structure if missing"""
```
bdraco, line 179:
```suggestion
        """Wrapper around access_tokens to automatically create underlaying data structure if missing."""
```

custom_components/unifiprotect/entity.py
```diff
@@ -99,8 +120,70 @@ def _async_updated_event(self) -> None:
...
+
+    @property
+    def extra_state_attributes(self) -> dict[str, Any]:
+        """Return the extra state attributes."""
+        attrs = super().extra_state_attributes or {}
+        return {
+            **attrs,
+            "access_token": self.access_tokens[-1],
+        }
+
```
bdraco, line 161:
```suggestion
```

Do we need the token separately since its already provided in the URL?

AngellusMortis: That was mirroring how the camera component did it. My assumption for that was it did it to ensure the access token was always in the state the in event of an unexpected shutdown/restart. The last one would not be lost.

bdraco: I'm pretty sure it was added for the frontend to consume it.  I don't think we need it for this case.

bdraco: https://github.com/home-assistant/frontend/blob/dev/src/data/camera.ts#L45

bdraco (approved):
Testing passed.

PR #383, opened 2021-12-07
Replaces all instances of Unifi

AngellusMortis:
Based on feedback from a PR I made to update the UniFi Network integration for HA and looking at other stuff. I renamed all instances of "Unifi" to "UniFi" in the code. This makes our code align more with Ubiquiti's brand.

For classes and such, I just removed "Unifi" all together. So `UnifiProtectData` becomes `ProtectData`. This both makes it easier to stay in-line with Ubiquiti's brand and shortens the classes for a number of classes we have in the code.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,7 +6,7 @@
     "configurations": [
         // Run "Start Home Assistant" task + `debugpy.start` HA service first
         {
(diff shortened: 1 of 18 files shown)


bdraco (approved):
Did some basic testing.  Nothing seems to have exploded 👍

PR #388, opened 2021-12-10
Refactors entity classes and setup

AngellusMortis:
I refactored and unified a lot of the patterns we were using for creating entities. 

* Combined `ProtectData` and `ProtectEntryData`
* There are now two base entity classes `ProtectDeviceEntity` (for adoptable devices) and `ProtectNVREntity` (for NVR entities)
   * This also combines the `ProtectApiClient` and `ProtectData` properties into a single one as well
* Added `async_all_device_entities` that allows generating Entities in a common pattern for platforms that need to support multiple different adoptable devices


AngellusMortis:
This should be "feature complete" so feel free to review. The only testing I have done so far is making sure HA loads with all of it, though. Working on that now.

briis:
Too tired today. Will test in the morning. 🍺

AngellusMortis:
All of my testing looks good.

custom_components/unifiprotect/binary_sensor.py
```diff
@@ -440,24 +344,52 @@ def _async_fire_motion_event(self) -> None:
...
+        description.name = (description.name or "{index}").format(index=index)
+        self._index = index
+        self.entity_description: ProtectBinaryEntityDescription = description
+        super().__init__(data, device)
+
+    @callback
+    def _async_update_device_from_protect(self) -> None:
+        super()._async_update_device_from_protect()
+
+        if self.entity_description.key.startswith(_KEY_DISK_HEALTH):
```
bdraco, line 370: You can reverse the condition and return early, then outdent below now that this is broken up.

bdraco (approved):
Tested on two production installs 👍

PR #389, opened 2021-12-10
Adds Detected Object sensor

### LINKED ISSUE #342: Extra Binary Sensor for Smart Detection
### New Feature

Is it possible to get an extra binary sensor with every camera when smart detection (person and or vehicle) is activated.

I have a ton of motion detection triggers, but it would be helpful if I could also receive the smart detection which is more filtered.

I used the:
- G4 Doorbell
- G4 Bullet

Thanks in advance

Naamloos

AngellusMortis:
* Fixes issue in `ThumbnailProxyView` from last refactor PR that was missed from linting

* `NEW`: Added Detected Object sensor for cameras with smart detections. Values are `none`, `person` or `vehicle`. Contains `event_score` and `event_thumb` attributes.
  * Closes #342

### DIFF
--- a/custom_components/unifiprotect/views.py
+++ b/custom_components/unifiprotect/views.py
@@ -12,6 +12,7 @@ from pyunifiprotect.api import ProtectApiClient
 from pyunifiprotect.exceptions import NvrError
 
 from .const import DOMAIN
+from .data import ProtectData
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


bdraco:
@AngellusMortis please fix conflicts.  This one requires more manual testing so I merged the other one first

AngellusMortis:
Just fixed it.

custom_components/unifiprotect/views.py
```diff
@@ -32,9 +33,10 @@ def _get_access_tokens(
         self, entity_id: str
     ) -> tuple[collections.deque, ProtectApiClient] | None:
 
-        for entry in self.data.values():
-            if entity_id in entry.protect_data.access_tokens:
-                return entry.protect_data.access_tokens[entity_id], entry.protect
+        entries: list[ProtectData] = list(self.data.values())
+        for entry in entries:
```
bdraco, line 37: Linear search probably doesn't matter here

AngellusMortis: Is there a better way to handle this? I do not think we can have a data store that persist outside of an entry, so iterating entries seems to be the only way to handle this.

bdraco: We'd have to make a `dict` index but its not worth it.

bdraco (approved):
Testing passed 👍

PR #391, opened 2021-12-10
Adds various new entities

AngellusMortis:
* Bumps pyunifiprotect to 1.3.0
* Adds `ufp_set_function` for switch entities
* Replaces various unit_of_measures with constants from HA

* `NEW`: Adds Paired Camera select entity for Viewports

* `NEW`: Adds "Received Data", "Transferred Data", "Oldest Recording", "Storage Used", and "Disk Write Rate" sensors for cameras. Disabled by default. Requires "Realtime metrics" option to update in realtime.

* `NEW`: (requires UniFi Protect 1.20.1) Adds "Voltage" sensor for doorbells. Disabled by default.

* `NEW`: Adds "System Sounds" switch for cameras with speakers

* `NEW`: Adds switches to toggle overlay information for video feeds on all cameras
 
* `NEW`: Adds switches to toggle smart detection types on cameras with smart detections

### DIFF
--- a/custom_components/unifiprotect/sensor.py
+++ b/custom_components/unifiprotect/sensor.py
@@ -12,10 +12,8 @@ from homeassistant.components.sensor import SensorEntity, SensorEntityDescriptio
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import (
     DATA_BYTES,
-    DATA_MEGABYTES,
     DATA_RATE_BYTES_PER_SECOND,
     DATA_RATE_MEGABITS_PER_SECOND,
-    DATA_RATE_MEGABYTES_PER_SECOND,
     DEVICE_CLASS_BATTERY,
     DEVICE_CLASS_HUMIDITY,
     DEVICE_CLASS_ILLUMINANCE,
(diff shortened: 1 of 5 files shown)


custom_components/unifiprotect/sensor.py
```diff
@@ -108,6 +115,58 @@ class ProtectSensorEntityDescription(ProtectRequiredKeysMixin, SensorEntityDescr
...
+        precision=2,
+    ),
+    ProtectSensorEntityDescription(
+        key="voltage",
+        name="Voltage",
+        device_class=DEVICE_CLASS_VOLTAGE,
+        entity_registry_enabled_default=False,
+        entity_category=ENTITY_CATEGORY_DIAGNOSTIC,
+        ufp_value="voltage",
+        # voltage will be null if device is not camera or not on 1.20.1+
```
bdraco, line 166: Can we avoid adding the sensor if its not supported?

bdraco: Maybe a min version required field?

AngellusMortis: That is what the `ufp_required_field` will do. `voltage` is `float | None`. If it evals to `None` (or 0.0) it will not be added.

AngellusMortis: I meant for that comment to be "why am I using `voltage` instead of `feature_flags.has_chime`.

bdraco (approved):
LGTM 👍 
Screen Shot 2021-12-10 at 14 23 47

PR #397, opened 2021-12-12
Adds CONF_OVERRIDE_CHOST and new dynamic doorbell service

### LINKED ISSUE #248: Allow FQDM in addition to IP Address
You can easily pull the IP address form a fully qualified domain name (FQDN) in your script, but restricting us to only using IP addresses eliminates everyone who has a dynamic IP address from the ISP.

In case you're not familiar, here is the easiest way to get the IP address form any FQDN

```

# this is what this feature request is all about.
# I want to use a domain name, like duckdns, so that the WAN IP is always up to date
$udmpro_fqdn=mydomain.com

#I would allow this parameter to be set, but if you're going to hard code it, use cloudflare
$dns_nameserver=dns1.cloudflare.com

# boom, easy grep of nslookup
ip_address="$(nslookup -type=A $udmpro_fqdn $dns_nameserver | grep "Address" | awk '{print $2}' | sed -n 2p)"
```

### LINKED ISSUE #396: Allow to set doorbell LCD text dynamically
### New Feature

As described in #263, I want to be able to set the LCD text value dynamically, which is no longer in v0.11, while it was before.

### Additional context

_No response_

AngellusMortis:
* `NEW`: Adds `unifiprotect.set_doorbell_message` service. This is just like the `unifiprotect.set_doorbell_lcd_message`, but it is not deprecated and it requires the Doorbell Text Select entity instead of the Camera entity. Should **only** be used to set dynamic doorbell text messages (i.e. setting the current outdoor temperate on your doorbell). If you want to use static custom messages, use the Doorbell Text Select entity and the `unifiprotect.add_doorbell_text` / `unifiprotect.remove_doorbell_text` service. `unifiprotect.set_doorbell_lcd_message` is still deprecated and will still be removed in the next release.
  * Closes #396

* `NEW`: Adds "Override Connection Host" config option. This will force your RTSP(S) connection IP address to be the same as everything else. Should only be used if you need to forcibly use a different IP address.
  * For sure closes #248

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## 0.11.0-dev (master)
 
+* `NEW`: Adds `unifiprotect.set_doorbell_message` service. This is just like the `unifiprotect.set_doorbell_lcd_message`, but it is not deprecated and it requires the Doorbell Text Select entity instead of the Camera entity. Should **only** be used to set dynamic doorbell text messages (i.e. setting the current outdoor temperate on your doorbell). If you want to use static custom messages, use the Doorbell Text Select entity and the `unifiprotect.add_doorbell_text` / `unifiprotect.remove_doorbell_text` service. `unifiprotect.set_doorbell_lcd_message` is still deprecated and will still be removed in the next release.
+  * Closes #396
+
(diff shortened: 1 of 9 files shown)


bdraco (approved):
LGTM.  Traveling so I could not test locally.

bdraco:
Finally got to a place where I could update my Home Assistant remotely. All looks good

PR #414, opened 2021-12-19
Cleans up deprecations

AngellusMortis:
* Removes all 0.11 deprecations
* Fixes pylint linting.


bdraco (approved):
Tested on both of my primary production systems.

No unexpected side effects observed

bdraco:
Thanks 👍

PR #415, opened 2021-12-20
Telegram notifications

AngellusMortis:


AngellusMortis:
@mjdyson The smart one should be done and working. Going to do motion and doorbell now. Stuff I changed:

* Updated the descriptions a bit to help with having more notification types
* Made it so the image being attached and "Silence" command are both optional based on your inputs
* Added a caption to the photo since the message/title params are not used when you attach a photo

mjdyson:
I like the approach so far! Hopefully you're able to work through the different callbacks :)

AngellusMortis:
@mjdyson I got all of the notification blueprints updated with Telegram. I also switched the response message for an actionable notification to use [answer callback query](https://www.home-assistant.io/integrations/telegram_bot#service-telegram_botanswer_callback_query) as well.

mjdyson:
All three seem to work and send a photo. Would it be possible to get a response from the silence action to say how long it's silenced for?

AngellusMortis:
It should. On Android it pops up as a toast notification in the top
Screenshot_20211220-121725500

mjdyson:
This looks great - great job, and very quickly done! 
The feedback notification works well - although i'm torn whether it would be better posted back in the chat window with an expiry time. Probably no need to include that in this PR. 
Would it be worth expanding the description text on cooldown to say something like "short cooldowns (<3s) may not be silenceable as the response time cant be met"?

AngellusMortis:
Updated all of the help text for them to hopefully clarify that.

AngellusMortis:
@mjdyson Everything looks good? If so I will go ahead and merge.

mjdyson:
Looks good to me. I'm sure others will give feedback if it needs a bit more work :)

PR #430, opened 2021-12-27
Adds 2022.2 upgrade warning

AngellusMortis:
Now that the first HA core PR has been merged and `unifiprotect` will officially be in 2022.2, I figured it is a good idea to add this warning for users. I will go through and add it to all of the recent releases/tags (probably back to 0.9.x versions) as well so hopefully no one will miss the noticed.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,6 +26,7 @@ jobs:
       - name: Install Linting Dependencies
         run: |
           git clone --depth=1 -b dev https://github.com/home-assistant/core.git hass
+          rm -rf ./hass/homeassistant/components/unifiprotect
           ln -s $GITHUB_WORKSPACE/custom_components/unifiprotect ./hass/homeassistant/components/unifiprotect
 
           python -m pip install --upgrade pip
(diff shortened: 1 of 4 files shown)


AngellusMortis:
@briis thoughts on the warning? The goal is to have both versions (HACS/HA Core) to either be the exact same for the 2022.2 release or at least be compatible with each other. So, if you have them both installed, it should not break your HA or `unifiprotect`.

briis:
I have no objections to the warning, I think you will find many people pleased that it will be part of core.

> if you have them both installed

As the domain is named `unifiprotect` on both of them, they will not be able to co-exist, so we need to add some text about how to upgrade once 2022.2 is out:

- Remove `unifiprotect` integration
- Delete the HACS installation
- Setup the core Integration

If we then want to test things outside the Core update cycle, the HACS version should be renamed to something else.

AngellusMortis:
I do not want to add official upgrade instructions yet as we have not actually tested it yet. We need to get all of the PRs merged so there is parity before we can test the migration process. 

> If we then want to test things outside the Core update cycle, the HACS version should be renamed to something else.

Nah, I think the HA release cycle is fast enough that should not be an issue.

Issue #431, opened 2021-12-27
Cannot use DNS host for connection host manually set within camera


tkk2112:
### Home Assistant Version?

2021.12.5

### UniFi Protect Version?

1.20.1

### UniFi Protect HA Integration version?

0.11.4

### What happened?

I upgrade both HA and the integration to latest, and then the integration wouldn't load, I didn't carefully check the logs so I don't know what the issues was. I delete the integration and reconfigured it with `se-gw.local` as host, that works fine in the wizard but the integration wont start saying `Failed to set up - Check the logs` the python exception raised is `ipaddress.AddressValueError: Expected 4 octets in 'se-gw.local'` which makes sense since I wront in a dns name instead of an IP address. But when deleting this integration and writing the IP directly its keeps saying `ipaddress.AddressValueError: Expected 4 octets in 'se-gw.local'`. Does it "cache" it or something?

### Relevant log output

```shell
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 313, in async_setup
    result = await component.async_setup_entry(hass, self)  # type: ignore
  File "/config/custom_components/unifiprotect/__init__.py", line 210, in async_setup_entry
    await data_service.async_setup()
  File "/config/custom_components/unifiprotect/data.py", line 74, in async_setup
    await self.async_refresh()
  File "/config/custom_components/unifiprotect/data.py", line 90, in async_refresh
    self._async_process_updates(await self.api.update(force=force))
  File "/usr/local/lib/python3.9/site-packages/pyunifiprotect/api.py", line 568, in update
    self._bootstrap = await self.get_bootstrap()
  File "/usr/local/lib/python3.9/site-packages/pyunifiprotect/api.py", line 717, in get_bootstrap
    return Bootstrap.from_unifi_dict(**data, api=self)
  File "/usr/local/lib/python3.9/site-packages/pyunifiprotect/data/base.py", line 100, in from_unifi_dict
    data = cls.unifi_dict_to_dict(data)
  File "/usr/local/lib/python3.9/site-packages/pyunifiprotect/data/bootstrap.py", line 132, in unifi_dict_to_dict
    return super().unifi_dict_to_dict(data)
  File "/usr/local/lib/python3.9/site-packages/pyunifiprotect/data/base.py", line 332, in unifi_dict_to_dict
    data[key] = cls._clean_protect_obj_dict(data[key], unifi_dicts[key], api)
[...]

PR #438, opened 2021-12-29
Backports various fixes from core merge

AngellusMortis:
There are no new code changes in this PR. Just copy/pastes from the HA core version. Not everything was ported due to risk bugs with the changes to the base entity class. Will likely make another PR after everything else is ported. Just wanted to get some of the major fixes here in. 

Backports fixes from Home Assistant core merge process:

* `FIX`: Improves relibility of entities when UniFi Protect goes offline and/or a device goes offline. Everything recovery seemlessly when UniFi Protect upgrades or firmware updates are applied (fixes https://github.com/briis/unifiprotect/issues/432).

* `FIX`: Improves relibility of `media_player` entities so they should report state better and be able to play longer audio clips.

* `FIX`: Fixes stopping in progress audio for `media_player` entities.

* `FIX`: Allows DNS hosts in addition to IP addresses (fixes https://github.com/briis/unifiprotect/issues/431).

* `FIX`: Fixes selection of default camera entity for when it is not the High Quality channel.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -4,7 +4,7 @@ RUN apt update \
     && sudo apt install -y libpcap-dev ffmpeg vim curl jq libturbojpeg0 \
     && mkdir -p /opt \
     && cd /opt \
-    && git clone --depth=1 -b dev https://github.com/home-assistant/core.git hass \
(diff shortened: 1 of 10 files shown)


bdraco (approved):
Reviewed for transcription errors only since this has already been reviewed on the core side

👍

PR #444, opened 2022-01-06
Adds generate_data service

AngellusMortis:
One last debug service to add. It was originally only a utility for generating data for tests, but it can be something we integrate into issue reporting

There is actually a bug in the anonymization of data, so right now only "actual" works. I would also like to make it so the log messages actually log instead of print to console if not ran in "CLI mode", but both things are upstream in the in `pyunifiprotect` so does not affect this PR or any code in it.

### DIFF
--- a/custom_components/unifiprotect/services.yaml
+++ b/custom_components/unifiprotect/services.yaml
@@ -21,7 +21,7 @@ take_sample:
           step: 10
           unit_of_measurement: seconds
     anonymize:
-      name:  Anonymize Data
+      name: Anonymize Data
       description: Data is anonymized by default to make safe for distribution.
       default: true
       required: true
--- a/custom_components/unifiprotect/services.yaml
+++ b/custom_components/unifiprotect/services.yaml
@@ -1,3 +1,32 @@
+generate_data:
+  name: Generate UniFi Protect Sample Data
+  description: Generates Sample Data from a given UniFi Protect instance. Useful for debugging or reporting issues.
+  fields:
+    device_id:
+      name: UniFi Protect NVR
+      description: Main NVR for the UniFi Protect instance you would like to target.
+      required: true
+      selector:
+        device:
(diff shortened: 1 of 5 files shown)


custom_components/unifiprotect/services.yaml
```diff
@@ -1,3 +1,32 @@
...
+      default: 30
+      required: true
+      selector:
+        number:
+          min: 0
+          max: 600
+          step: 10
+          unit_of_measurement: seconds
+    actual:
+      name:  Actual Data?
```
bdraco, line 24:
```suggestion
    anonymize:
      name: Anonymize Data
```

custom_components/unifiprotect/services.yaml
```diff
@@ -1,3 +1,32 @@
...
+      description: Generate actual data from UniFi Protect. Otherwise data is automatically anonymized to make safe for distribution.
+      default: false
```
bdraco, line 26:
```suggestion
      default: true
```

custom_components/unifiprotect/services.yaml
```diff
@@ -1,3 +1,32 @@
+generate_data:
```
bdraco, line 1:
```suggestion
take_sample:
```

Maybe this is a bit more descriptive ?

custom_components/unifiprotect/services.yaml
```diff
@@ -20,10 +20,10 @@ generate_data:
           max: 600
           step: 10
           unit_of_measurement: seconds
-    actual:
-      name:  Actual Data?
-      description: Generate actual data from UniFi Protect. Otherwise data is automatically anonymized to make safe for distribution.
-      default: false
+    anonymize:
+      name:  Anonymize Data
```
bdraco, line 24:
```suggestion
      name: Anonymize Data
```

bdraco (approved):

Issue #440, opened 2022-01-03
Sensor Spelling


crlogic:
### Home Assistant Version?

2021.12.7

### UniFi Protect Version?

1.21

### UniFi Protect HA Integration version?

0.12

### What happened?

Check line 309 in  https://github.com/briis/unifiprotect/blob/master/custom_components/unifiprotect/sensor.py

```
name="Recording Capcity",
```

PR #449, opened 2022-01-08
Batch of small bugfixes

AngellusMortis:
* Fixes memory sensor if no data is reported
* Fixes https://github.com/briis/unifiprotect/issues/440
* Fixes is connected check for cameras
* Adds last_trip_time back to camera motion entity


bdraco (approved):
LGTM. Nothing obvious broke in manual testing

